PR #12, opened 2016-03-03
Added IO Data, altered environment data container ID

Cerfoglg:


database_schemas/cassandra/benchflow.cql
```diff
@@ -23,14 +23,26 @@ CREATE TABLE database_sizes (
...
   database_name text,
-  database_table text,
   size bigint,
   PRIMARY KEY ((trial_id, experiment_id), dbms)
 );
 
+CREATE TABLE io_data (
+  experiment_id text,
+  trial_id text,
+  container_properties_id uuid,
```
VincenzoFerme: The name is now `environment_data_id`

database_schemas/cassandra/benchflow.cql
```diff
@@ -23,14 +23,26 @@ CREATE TABLE database_sizes (
...
+  sync bigint,
+  async bigint,
+  total bigint,
+  PRIMARY KEY ((trial_id, experiment_id), container_properties_id)
+);
+
 CREATE TABLE environment_data (
   environment_data_id uuid,
-  container_properties_id uuid,
+  container_id text,
```
VincenzoFerme: This should be `container_properties_id` because the only unique id we have is the one we generate. It is possible, although unlikely, that in the same `experiment` and `trial` we have two containers with the same `container_id`.

database_schemas/cassandra/benchflow.cql
```diff
@@ -23,14 +23,26 @@ CREATE TABLE database_sizes (
   trial_id text,
   dbms text,
   database_name text,
-  database_table text,
   size bigint,
   PRIMARY KEY ((trial_id, experiment_id), dbms)
 );
 
+CREATE TABLE io_data (
```
VincenzoFerme: The table should have an `io_data_id uuid` field.

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme The full id of a container is very long, the likelihood of a conflict, especially considering we also have experiment id and trial id tied to them, is very unlikely. If I don't use the container id provided by docker and instead try generating my own, it will be really difficult to track considering the setup with collectors and Spark Tasks Sender messaging

database_schemas/cassandra/benchflow.cql
```diff
@@ -41,7 +53,7 @@ CREATE TABLE environment_data (
   network_interfaces map<text, uuid>,
   experiment_id text,
```
VincenzoFerme: Lets add an `io_data uuid` to map the data in the `io_data` table.

database_schemas/cassandra/benchflow.cql
VincenzoFerme: We still might have a conflict, and it is not affordable.

Why is it difficult?

- You get the ID generated by Docker and pass it to the transformer
- The transformer uses it to get the right file, with the `container_properties` data
- You store the container properties in the `container_properties` data and get back the `uuid` generated for/from Cassandra
- You then use this new generated `uuid` for everything else you are already doing in the transformer
- You submit the analysers on each of the generated `uuid`s or one analyser working on all the containers for a given experiment and/or trial.

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme But aren't they cumulative the way they are collected? So it's only one row per trial and container

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme Then I'll need the stats collector to collect the container properties and save them on Minio so the stats transformer can do all that, because the spark tasks sender doesn't have dependencies for transformers. I'll need to look into this.

VincenzoFerme: Yes, either we have a dedicated collector to collect the discussed information, or we rely on one of the collectors that interact with the Docker socket already, and currently it is only the stats collector. Moreover I also think that the fact that the stats collector, also collects the metadata for which it collects the statistics, it is consistent. We might change our mind at some point, but currently I think it is a good idea in the system we are building.

PR #13, opened 2016-03-04
Adjusted float types for data

Cerfoglg:


database_schemas/cassandra/benchflow.cql
```diff
@@ -184,24 +183,24 @@ CREATE TABLE host_properties (
...
-  process_duration_ci095_max float,
-  process_duration_median_min float,
-  process_duration_median_max float,
-  process_duration_mode_min float,
-  process_duration_mode_max float,
-  process_duration_min float,
-  process_duration_max float,
-  process_duration_sd float,
+  process_duration_weighted_avg double,
+  process_duration_q1_min bigint,
```
VincenzoFerme: The process duration can be also decimal, it actually depends on the precision of the system under test.

database_schemas/cassandra/benchflow.cql
```diff
@@ -210,128 +209,175 @@ CREATE TABLE exp_process_duration (
...
+  cpu_mode_min double,
+  cpu_mode_max double,
+  cpu_min double,
+  cpu_max double,
+  cpu_sd double,
   cpu_best list<text>,
   cpu_average list<text>,
   cpu_worst list<text>,
   experiment_id text,
+  container_id text,
```
VincenzoFerme: Should we use `container_id` as group/clustering key

VincenzoFerme: ?

database_schemas/cassandra/benchflow.cql
```diff
@@ -210,128 +209,175 @@ CREATE TABLE exp_process_duration (
...
-  ram_ci095_max float,
-  ram_median_min float,
-  ram_median_max float,
-  ram_mode_min float,
-  ram_mode_max float,
-  ram_min float,
-  ram_max float,
-  ram_sd float,
+  ram_weighted_avg double,
+  ram_q1_min bigint,
```
VincenzoFerme: I guess RUM has decimal positions.

database_schemas/cassandra/benchflow.cql
```diff
@@ -210,128 +209,175 @@ CREATE TABLE exp_process_duration (
...
+  ram_mode_min bigint,
+  ram_mode_max bigint,
+  ram_min bigint,
+  ram_max bigint,
+  ram_sd double,
   ram_best list<text>,
   ram_average list<text>,
   ram_worst list<text>,
   experiment_id text,
+  container_id text,
```
VincenzoFerme: Should we use `container_id` as group/clustering key?

database_schemas/cassandra/benchflow.cql
```diff
@@ -210,128 +209,175 @@ CREATE TABLE exp_process_duration (
...
+  total_q3 bigint,
+  total_p95 bigint,
+  total_ci095_min double,
+  total_ci095_max double,
+  total_me double,
+  total_median bigint,
+  total_mode list<bigint>,
+  total_sd double,
+  experiment_id text,
+  device text,
```
VincenzoFerme: This should be the `io_data_id` of the `io_data` table.

database_schemas/cassandra/benchflow.cql
```diff
@@ -210,128 +209,175 @@ CREATE TABLE exp_process_duration (
...
-  process_duration_ci095_min float,
-  process_duration_ci095_max float,
-  process_duration_me float,
-  process_duration_median float,
-  process_duration_mode list<float>,
-  process_duration_min float,
-  process_duration_max float,
-  process_duration_sd float,
+  process_duration_avg double,
+  process_duration_q1 bigint,
```
VincenzoFerme: The process duration can be also decimal, it actually depends on the precision of the system under test.

database_schemas/cassandra/benchflow.cql
```diff
@@ -339,50 +385,52 @@ CREATE TABLE trial_process_duration (
...
-  ram_ci095_max float,
-  ram_me float,
-  ram_median float,
-  ram_mode list<float>,
-  ram_min float,
-  ram_max float,
-  ram_sd float,
-  ram_integral float,
+  ram_avg double,
+  ram_q1 bigint,
```
VincenzoFerme: I guess RAM has decimal positions. Or at least, it should.

database_schemas/cassandra/benchflow.cql
```diff
@@ -401,4 +449,15 @@ CREATE TABLE trial_byte_size (
   experiment_id text,
   trial_id text,
   PRIMARY KEY (experiment_id, trial_id)
+);
+
+CREATE TABLE trial_IO (
```
VincenzoFerme: Use lower case `io`

database_schemas/cassandra/benchflow.cql
```diff
@@ -401,4 +449,15 @@ CREATE TABLE trial_byte_size (
...
+  reads bigint,
+  writes bigint,
+  total bigint,
+  experiment_id text,
+  trial_id text,
+  container_id text,
+  device text,
```
VincenzoFerme: This should be the `io_data_id` of the `io_data` table.

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme Turns out decimals need special treatment for the cassandra connector. Double will work fine for us for now (double offers a lot of precision anyways). I might open an issue regarding this.

VincenzoFerme: A wanted to say `double`. Basically it can have decimal precision, `bigint` it is not suitable for that.

database_schemas/cassandra/benchflow.cql
Cerfoglg: We are collecting the RAM as integers, so for things like the max and min for instance we can use bigint.

database_schemas/cassandra/benchflow.cql
Cerfoglg: the io_data contains many rows, with io_data_id identifying single rows. If anything we need the container_id, since these data are for a certain container

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme Got it, then that can be changed no problem

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme the io_data_ids from io_data now refer to single collected points, like in environment_data , so no point storing it since the trial analyser now finds the max for all values per device.

VincenzoFerme: Right, ok. Ok, thank you.

database_schemas/cassandra/benchflow.cql
VincenzoFerme: @Cerfoglg on the following link it is stated that 

> All memory amounts are in bytes (except for event counters).

It means that, when we divide `/1024/1024` to get the amount in MB, we might end up with decimal numbers.

database_schemas/cassandra/benchflow.cql
VincenzoFerme: Right, lets use `container_id`

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme I didn't see that one. I don't think I've ever seen decimals in all the data we have collected for memory, but better be safe then, I'll use double here as well

database_schemas/cassandra/benchflow.cql
```diff
@@ -339,50 +386,52 @@ CREATE TABLE trial_process_duration (
...
+  cpu_mode list<double>,
+  cpu_min double,
+  cpu_max double,
+  cpu_sd double,
+  cpu_integral double,
   experiment_id text,
   trial_id text,
-  PRIMARY KEY (experiment_id, trial_id)
+  container_id text,
+  PRIMARY KEY (experiment_id, container_id, trial_id)
```
VincenzoFerme: Why first the `container_id` and the the `trial_id`?

database_schemas/cassandra/benchflow.cql
```diff
@@ -339,50 +386,52 @@ CREATE TABLE trial_process_duration (
...
+  ram_mode list<bigint>,
+  ram_min bigint,
+  ram_max bigint,
+  ram_sd double,
+  ram_integral double,
   experiment_id text,
   trial_id text,
-  PRIMARY KEY (experiment_id, trial_id)
+  container_id text,
+  PRIMARY KEY (experiment_id, container_id, trial_id)
```
VincenzoFerme: Why first the `container_id` and the the `trial_id`?

database_schemas/cassandra/benchflow.cql
```diff
@@ -401,4 +450,15 @@ CREATE TABLE trial_byte_size (
...
+
+CREATE TABLE trial_io (
+  reads bigint,
+  writes bigint,
+  total bigint,
+  experiment_id text,
+  trial_id text,
+  container_id text,
+  device text,
+  PRIMARY KEY (experiment_id, container_id, trial_id, device)
```
VincenzoFerme: Why first the `container_id` and the the `trial_id`?

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme When you perform a query with where clauses with Cassandra for some reason you can't specify a column value if you didn;t specify all previous ones as well in the primary key, meaning that if I had trial_id first I couldn't have a query with "where experiment =x and container_id=y", because it will give an error saying that I needed to specify trial_id as well if I wanted to specify container_id. It's part of how Cassandra does clustering.

database_schemas/cassandra/benchflow.cql
VincenzoFerme: You never saw it because Docker stats provide you these data using the amount of bytes, but then you perform a conversion to MB. And by the way, I suspect the following two conversion should be performed with a `double` precision: https://github.com/benchflow/data-transformers/blob/dev/data-transformers/transformers/statsTransformer.py#L71

database_schemas/cassandra/benchflow.cql
VincenzoFerme: Ok, thank you for the explanation.

PR #16, opened 2016-03-08
Added table to store cpu per core metrics

Cerfoglg:


database_schemas/cassandra/benchflow.cql
```diff
@@ -46,6 +46,7 @@ CREATE TABLE environment_data (
   container_id text,
   read_time text,
   cpu_percpu_usage list<bigint>,
+  cpu_percpu_percent_usage list<double>,
```
VincenzoFerme: Why are these percentage of the `double` type, while the `cpu_percent_usage` is a `float`?

database_schemas/cassandra/benchflow.cql
```diff
@@ -233,6 +234,32 @@ CREATE TABLE exp_cpu (
   PRIMARY KEY (experiment_id, container_id)
 );
 
+CREATE TABLE exp_cpu_core (
```
VincenzoFerme: You can think about having a list of value for each field. At the end we are always interested to look at these data all at once, so an ordered list of data points should be suitable.

database_schemas/cassandra/benchflow.cql
```diff
@@ -233,6 +234,32 @@ CREATE TABLE exp_cpu (
...
+  cpu_median_max double,
+  cpu_min double,
+  cpu_max double,
+  cpu_sd double,
+  cpu_best list<text>,
+  cpu_average list<text>,
+  cpu_worst list<text>,
+  experiment_id text,
+  container_id text,
+  core bigint,
```
VincenzoFerme: Why do we need a `bigint` for the number of core?

database_schemas/cassandra/benchflow.cql
```diff
@@ -233,6 +234,32 @@ CREATE TABLE exp_cpu (
...
+  cpu_p95_min double,
+  cpu_p95_max double,
+  cpu_ci095_min double,
+  cpu_ci095_max double,
+  cpu_median_min double,
+  cpu_median_max double,
+  cpu_min double,
+  cpu_max double,
+  cpu_sd double,
+  cpu_best list<text>,
```
VincenzoFerme: We can remove `cpu_best`, `cpu_average` and `cpu_worst`

database_schemas/cassandra/benchflow.cql
```diff
@@ -404,6 +431,29 @@ CREATE TABLE trial_cpu (
   PRIMARY KEY (experiment_id, container_id, trial_id)
 );
 
+CREATE TABLE trial_cpu_core (
```
VincenzoFerme: The same as the `exp_cpu_core`

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme I missed that when revising the table. Fixing now

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme That would mean having rows which is essentially nothing but lists for almost all columns. Not sure if this is the best way to do this. Having only one or two columns being lists is one thing, but all of them might not be the best for Cassandra.

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme Maybe I was thinking too high. I don;t know how many cores there could potentially be, so I did that.

database_schemas/cassandra/benchflow.cql
VincenzoFerme: Even if you always want to access to that data at once?

database_schemas/cassandra/benchflow.cql
VincenzoFerme: An `int` should be fine

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme I looked into this again, and I think unless the lists become gigantic (which is unlikely) then yes, we are safe using lists instead of having one row per core.

database_schemas/cassandra/benchflow.cql
Cerfoglg: @VincenzoFerme okay

PR #111, opened 2017-03-24
Updates to use separate objects for read/write parsing

### LINKED ISSUE #58: Use separate objects for read/write parsing
Use separate objects to read/write parsing so that we can make use of `Try`s for the read, and avoid them for the write.

Also update the code of the write according to example in [BenchFlowTestYamlProtocol.scala](https://github.com/benchflow/benchflow/blob/feature-dsl-v1-initial/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/BenchFlowTestYamlProtocol.scala#L95).

jeppe-style:
Refers to issue (closes https://github.com/benchflow/benchflow/issues/58).

@igormoreno I need your help for solving [this](https://github.com/benchflow/benchflow/pull/111/files#diff-5835d0a3fe4a663cb3e6cbfbb7ccb49cR81) -> thanks for your help!


VincenzoFerme:
@jeppe-style ping me when it is in a state it can be reviewed, also after checking at https://www.codacy.com/app/BenchFlow/benchflow/pullRequest?prid=575554

jeppe-style:
@VincenzoFerme now it is in a state to be reviewed

VincenzoFerme:
@jeppe-style since you also committed many code changes related to the style checker, can you please point us to the actual commit where there is business logic to review? Thank you.

jeppe-style:
@igormoreno I still get some `user correct indentation` and some `file must end with newline character` after running scalariform using maven. (All warnings under `dsl/definition/simone/` can be ignored because we will delete or move these files. Also files under `dsl/deployment/` we can ignore in this pull-request since I have not worked on those files and also the ones under `/src/test/`.)

@VincenzoFerme the non-codestyle changes are the following:
- in all *YamlProtocol.scala files there is now *ReadFormat and *WriteFormat objects. This is so we in the read can include Try, but can avoid it in the write since we know there are no type errors. Since we have to implement both read and write in both of these objects we have added methods that throws `UnsupportedOperationExceptions`:
  - example [*YamlProtocol](https://github.com/benchflow/benchflow/pull/111/files#diff-5e7781bad0e954fc7ef0263fcf621e46R78)
  - [definition of exception methods](https://github.com/benchflow/benchflow/pull/111/files#diff-c65915cb06a22e7fa427601221735eb2R35)
- in the [MixYamlProtocol.scala](https://github.com/benchflow/benchflow/pull/111/files#diff-5835d0a3fe4a663cb3e6cbfbb7ccb49cR71) file we completed the write method to support all possible MixTypes

VincenzoFerme (approved):
After @igormoreno gives you review and help you fix what you report in https://github.com/benchflow/benchflow/pull/111#issuecomment-289150055, ping me so I can merge.

igormoreno:
I know you added that:
```
// TODO - add documentation reference with reasons why we separate read and write into 2 objects
```
referring to the conversation we've had to document this design decision. But in general I would advise to have this kind of documentation committed in the same pull request. Especially regarding important design decisions. For example, at the time we considered 3 possible courses of action and it would be nice to have it written there (and I suppose you could use it for your thesis). It doesn't have to be a long documentation. My point is that the higher the "temporal gap" between the changes and the documentation the harder it is to keep track I think, both in terms of trying to track the project in general but also one starts to forget details and its harder to document. Of course you don't want to go crazy on this but I think it's important to do it for the most important pieces of documentation (which often have subtle details in them) (and of course there are some subjective evaluation involved in deciding what is "most important")

igormoreno:
Regarding the warnings on indentation:
Here’s the deal. I looked for it a lot. The issue is that scalastyle checks that multiline constructor arguments are indented with 4 spaces instead of 2 (as written in the scala style guideline), but scalariform doesn’t format that correctly: https://github.com/scala-ide/scalariform/pull/227 https://github.com/scala-ide/scalariform/pull/192 (these are 2 pending pull requests that should solve this issue, but the last activity on them seem to be in September of last year). Another option would be to fine tune scalastyle to not complain about this specific issue but even though there are some configuration available in terms of indentation (for example for multiline method parameter) there seems to be no configuration for multiline constructor parameter. So my suggestion is that we turn off the indentation checks. If we compile and test the code always using maven the code will be always automatically formatted by scalariform anyway (because it runs in a previous phase in maven) so the indentation will be okay.

I'm going to add to this pull request this change

igormoreno (approved):

jeppe-style:
@VincenzoFerme I added now a brief description of the design decision @igormoreno felt was missing in his comment https://github.com/benchflow/benchflow/pull/111#issuecomment-290009104

The remaining code check errors are due to files that are not part of the changes in this commit. For me we can merge now.

PR #193, opened 2017-04-18
Ports the Driver-Maker to the Single Repo

VincenzoFerme:
@jeppe-style please review that the porting is fine and does not include artefacts that should be removed


benchflow-driver-maker/.travis.yml
```diff
@@ -0,0 +1,59 @@
+sudo: required
```
jeppe-style: I don't think this file is necessary since we use wercker.

benchflow-driver-maker/api/.gitignore
```diff
@@ -0,0 +1 @@
+/target/
```
jeppe-style, line 1: I think we can cover this by .gitignore in the root of the repo or project, so that we don't have so many .gitignore files.

VincenzoFerme:
@igormoreno code checker configuration scheduled in #242
@vincenzoferme DevOps support scheduled in #243

jeppe-style (approved):

PR #225, opened 2017-04-18
Feature dsl experiment generation with number of users parameter

jeppe-style:


VincenzoFerme:
@jeppe-style fix https://www.codacy.com/app/BenchFlow/benchflow/pullRequest?prid=612662, or @igormoreno improve the settings if these warnings are ok to have. Thanks!

jeppe-style:
Yes, I have changes that will solve these issues but I thought to do it in a second pull-request, basically I will remove the Try and throw an exception instead since that is easier to handle in Java anyway.

jeppe-style:
@VincenzoFerme the code is now updated so that the checks pass and now it will be easier to use in Java.

@igormoreno  Some of the issues that are shown in codacy I don't get when running `mvn verify`.

benchflow-dsl/pom.xml
```diff
@@ -36,13 +38,13 @@
 
         <dependency>
             <groupId>org.scala-lang.modules</groupId>
-            <artifactId>scala-xml_2.11</artifactId>
+            <artifactId>scala-xml_2.12</artifactId>
```
VincenzoFerme: `2.12` can also be defined as properties, or probably can be even extracted from `<scala.xml.version>` by filtering out the last `.1`. This applies to all the cases in this, and other files.

benchflow-dsl/src/test/resources/tests/exploration/users/benchflow-test.yml
```diff
@@ -0,0 +1,90 @@
...
+  goal:
+    type: load
+
+    exploration_space:
+      workload:
+        users:
+          values: [5, 10, 15, 20]
+
+
+  users: 10
```
VincenzoFerme: Why do you have the setting also here, if you define an exploration space on users?

VincenzoFerme (changes requested):
This is just a very quick review, we should plan more in depth reviews in next iterations. An issue has been opened to schedule a complete code review (#244)

igormoreno (approved):
- Consider using string interpolation if you need to do 
too many string concatenations. I committed a small example.

- It would be nice to try to get read of those hardcoded strings: "configuration.goal.exploration_space.workload.users.". Finding your way somehow to build them dynamically. But it looks like that's not something trivial. So I would leave it for another iteration

igormoreno:
Issue regarding hardcoded strings: #249

jeppe-style:
Thanks @igormoreno and @VincenzoFerme for your comments! I have now included/fixed them so we should be good to go.

VincenzoFerme (approved):

PR #226, opened 2017-04-18
Relies one the WERCKER_OUTPUT_DIR to speed up the Wercker store step

VincenzoFerme:
Relies one the `WERCKER_OUTPUT_DIR` to speed up the Wercker store step


jeppe-style (approved):
Seems fine to me.

PR #228, opened 2017-04-19
Data manager migration to dropwizard and automatic stylechecker and formatter

igormoreno:
Main points:

---
About the migration from Play framework to Dropwizard:

It consists mostly of:

- Migration from sbt to maven. This includes:
  - Support for running tests using scalatest
  - Support for incremental and continuos compilation and testing
  - Proper merging of akka configuration into fat jar
- Translation of controllers
- Translation of integration tests
- Simplifying dependences (for example in the case of logging)
- Changing data serialization from (json -> binary) to (json -> text)
  This will allow any application to restore the data if necessary

Still pending:

- Migration of the configuration system
- Proper dependency injection
- Document key steps of the migration

---
About the automatic style guideline checker and the automatic formatter:

Adding to maven a java style guideline checker:
https://maven.apache.org/plugins/maven-checkstyle-plugin/

It is set up to follow the google java style guidelines:
https://google.github.io/styleguide/javaguide.html

You can run it in isolation with
```
mvn checkstyle:check
```

In terms of the maven development lifecycle phases, these static
analysis tools are typically bound to the phase `validate` (which
happens before `compile`) or the phase `verify` (which happens after
`package`). I've looked online at several discussions on this topic. I
don't think neither of these choices are good. Using the `validate`
phase is bad because i think the 1st static check should be performed by
the compiler and if we use the `verify` phase we have to wait for the
entire packaging phase to happen and that takes quite some time. So i
decided to bind all the static analysis tools to the `test` phase.
```
mvn test
```

I'm working on adding other static checks that are also available in the
cadacy platform.

I also added to maven a java code formatter also following the google
java style guidelines. It will run automatically before compilation.
Formatting configuration files for IDEs are also
available: https://github.com/coveo/fmt-maven-plugin

I am using maven configuration inheritance to share the configuration of
these and other plug-ins among multiple projects


igormoreno:
I can't make codacy pickup the settings from our configuration file. They say:
"If you have configuration files for your static analysis tool of choice, you can add it to your repository (and Codacy will pick it up automatically)."
https://support.codacy.com/hc/en-us/articles/207994335-Code-Patterns

 we have a file right there!
https://github.com/benchflow/benchflow/blob/devel/scalastyle-config.xml

But it doesn't work :-(

VincenzoFerme:
@igormoreno ping us when all it completed and working with code checkers so we can review it. What you write in https://github.com/benchflow/benchflow/pull/228#issue-222640932, should also be part of the documentation.

VincenzoFerme:
Closed in favour of #237 and #238

PR #235, opened 2017-04-19
Adds project for common minio client

jeppe-style:
@VincenzoFerme can you please check if we need the `maven-shade-plugin`. Right now it is commented in the `pom.xml`.

The project does not contain any devops related configurations at the moment.


VincenzoFerme:
@jeppe-style the `maven-shade-plugin` is needed to create a fat jar and solve all the possible dependencies conflicts so that we have the application and its dependencies in a single jar.

jeppe-style:
@VincenzoFerme ok - so we should always include that in all projects?

jeppe-style:
Reminder to also add the method as described in this issue https://github.com/benchflow/benchflow/issues/245

igormoreno (approved):

igormoreno:
This pull request got too big. We need to merge it and follow up with other requests.

Related issues:
- #245 
- #259 
- #262 
- #263 
- #264

PR #237, opened 2017-04-19
Java Style Checker and Formatter

igormoreno:
Okay guys. I decided to separate these things in a different feature branch.
Here's what you need to know:

---
Adding to maven a java style guideline checker:
https://maven.apache.org/plugins/maven-checkstyle-plugin/

It is set up to follow the google java style guidelines:
https://google.github.io/styleguide/javaguide.html

You can run it in isolation with
```
mvn checkstyle:check
```

In terms of the maven development lifecycle phases, these static
analysis tools are typically bound to the phase `validate` (which
happens before `compile`) or the phase `verify` (which happens after
`package`). I've looked online at several discussions on this topic. I
don't think neither of these choices are good. Using the `validate`
phase is bad because i think the 1st static check should be performed by
the compiler and if we use the `verify` phase we have to wait for the
entire packaging phase to happen and that takes quite some time. So i
decided to bind all the static analysis tools to the `test` phase.
```
mvn test
```

I'm working on adding other static checks that are also available in the
cadacy platform.

I also added to maven a java code formatter also following the google
java style guidelines. It will run automatically before compilation.
Formatting configuration files for IDEs are also
available: https://github.com/coveo/fmt-maven-plugin

I am using maven configuration inheritance to share the configuration of
these and other plug-ins among multiple projects

---

in addition I want to share with you guys a few things I discovered:

- One of the main usage examples in the checkstyle website is wrong!
https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html
the _configuration_ section should be *outside* the _executions_ section. Not inside. It took me ages to find out why my configuration was not working :-(

- The formatter i'm using is a maven plug-in which basically calls the  software Google users to format their Java code:
https://github.com/google/google-java-format
which they say doesn't accept configuration on purpose to make sure nobody deviates from the standard. If we decide we want to have a standard that is based on the Google standard but with something different we have to use a different formatter. For example:
http://code.revelc.net/formatter-maven-plugin/formatter-maven-plugin/index.html
which accepts configuration.


igormoreno:
Part of the original documentation marked as migrated:
https://github.com/VincenzoFerme/benchflow-temporary-documentation/blob/master/development/contributing-to-benchflow.md

VincenzoFerme (approved):

igormoreno:
Why were the other commits not merged?

PR #238, opened 2017-04-19
Data-manager Migration from Play Framework to Dropwizard

igormoreno:
The migration consists mostly of:

- Migration from sbt to maven. This includes:
  - Support for running tests using scalatest
  - Support for incremental and continuos compilation and testing
  - Proper merging of akka configuration into fat jar
- Translation of controllers
- Translation of integration tests
- Simplifying dependences (for example in the case of logging)
- Changing data serialization from (json -> binary) to (json -> text)
  This will allow any application to restore the data if necessary

Still pending:

- [X] Migration of the configuration system
- [x] Proper dependency injection
- [ ] https://benchflow.slack.com/archives/C5D3KM6GK/p1495128171685736 (configuration.yml, GoogleDriveFactory)
- [ ] Document key steps of the migration
- [ ] Align Maven build options as in https://github.com/benchflow/benchflow/pull/495/


igormoreno:
The most important pending items were added to the following issues:
- #278 
- #277 
- #292 
- #293

VincenzoFerme:
@igormoreno I enabled CI build and Docker Image deployment for the service. The build fails, I guess the Test compliance described in https://github.com/benchflow/benchflow/issues/293 has to be applied before being able to merge the `data-manager`

jeppe-style:
Suggest to comply with API standard as defined in https://github.com/benchflow/benchflow/issues/293

Why do we have the package `service`? We don't have that in the other services.

Missing test that actually starts the application (see example in test-manager https://github.com/benchflow/benchflow/blob/devel/benchflow-test-manager/application/src/test/java/cloud/benchflow/testmanager/BenchFlowTestManagerApplicationIT.java)

Not all dependencies have the version set as a property.

Maybe also to add Swagger (see example in test-manager https://github.com/benchflow/benchflow/blob/devel/benchflow-test-manager/pom.xml#L130)

Probably also dropwizard-template-config is needed to read environment variables when running in docker (see example in test-manager https://github.com/benchflow/benchflow/blob/devel/benchflow-test-manager/pom.xml#L90)

benchflow-data-manager/src/main/java/cloud/benchflow/datamanager/service/DataManagerConfiguration.java
```diff
@@ -0,0 +1,33 @@
+package cloud.benchflow.datamanager.service;
```
jeppe-style: should probably go into package `configurations` like the other services

igormoreno:
Where are these standards I have to comply with?

benchflow-data-manager/src/main/java/cloud/benchflow/datamanager/service/DataManagerConfiguration.java
igormoreno: Sure. That's part of #293

igormoreno:
To do:

- [x] folder structure: #293 
- [ ] URL standards: #349 
- [x] configuration files: #277 
- [ ] integration tests: #348 
- [x] add swagger: #350
- [x] dependency version in properties: #351

igormoreno:
I'm getting and error in wercker:
```
  Cause: java.lang.IllegalStateException: Could not find docker-compose, looked in: DockerCommandLocations{possiblePaths=[null, /usr/local/bin/docker-compose, /usr/bin/docker-compose]}
```

but i'm using "docker-compose-rule" which is also used in other projects. why doesn't it work?

VincenzoFerme:
Look in the following part of the `wrecker.yml` for the settings you need for Wrecker: https://github.com/benchflow/benchflow/blob/devel/wercker.yml#L114. In Wercker you can not interact with Docker Compose, so i came up with a flexible alternative. The code you took from @jeppe-style has also been modified by me to adapt it to Wercker.

PR #194, opened 2017-04-18
Includes files from benchflow/experiments-manager repo

jeppe-style:

PR #241, opened 2017-04-20
Adds experiment lifecycle to Experiment Manager

jeppe-style:
This pull-request changes the lifecycle from one task into multiple and adds a controller to manage the  transitions.

Once https://github.com/benchflow/benchflow/pull/194 has been accepted we can review this.


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/demo/DriversMakerCompatibleID.java
```diff
@@ -0,0 +1,40 @@
+package cloud.benchflow.experimentmanager.demo;
```
VincenzoFerme, line 1: Why do you have this demo file on this branch?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/experiment/ExperimentRunningTask.java
```diff
@@ -0,0 +1,185 @@
...
+import static cloud.benchflow.experimentmanager.constants.BenchFlowConstants.MODEL_ID_DELIMITER;
+import static cloud.benchflow.faban.client.responses.RunStatus.Code.*;
+
+/**
+ * @author Jesper Findahl (jesper.findahl@usi.ch)
+ *         created on 2017-04-19
+ */
+public class ExperimentRunningTask extends CancellableTask {
+
+    private static final String TEMP_DIR = "./tmp";
```
VincenzoFerme, line 42: I would use a different subdirectory of the `tmp` folder for each of the classes needing it, maybe named using the lowercase name of the class. This to reduce possible conflicts in writing concurrently to the disk by many classes.

VincenzoFerme (changes requested):
A complete code review is scheduled in https://github.com/benchflow/benchflow/issues/244, but here I mention something we should update.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/demo/DriversMakerCompatibleID.java
jeppe-style: This is because as long as the driver-maker is not updated we need to use this class. Since other services are behind it is very difficult to work with changes in different branches. At least in this way it should work with the running system.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/experiment/ExperimentRunningTask.java
jeppe-style: If you look at the usage at [line 101](https://github.com/benchflow/benchflow/pull/241/files/20a81db7f54fbd40836063bbd0b09402dd338bae#diff-343ab7ec46b9117a2b4e7ed28229f9a6R101) you can see that in addition to this there is a unique path based on the experimentID, so I think it should be fine.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/demo/DriversMakerCompatibleID.java
VincenzoFerme: For me it is all right, just make it explicit in the javadoc of the class.

jeppe-style: Javadoc has been added and this is mentioned in the following issue https://github.com/benchflow/benchflow/issues/260

Issue #279, opened 2017-05-04
Can't build projects with local dependences


igormoreno:
On the experiment-manager:
```
[ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.3.0:wget (download-faban-client) on project benchflow-experiment-manager: IO Error: Could not get content -> [Help 1]
```

on the drivers-maker:
```
[ERROR] Failed to execute goal on project application: Could not resolve dependencies for project cloud.benchflow.drivers-maker:application:jar:0.1.0: The following artifacts could not be resolved: org.apache:httpclient:jar:3.1, cloud.benchflow:faban-driver:jar:0.1, cloud.benchflow:faban-agents:jar:0.1: Failure to find org.apache:httpclient:jar:3.1 in https://jitpack.io was cached in the local repository, resolution will not be reattempted until the update interval of jitpack.io has elapsed or updates are forced -> [Help 1]
```

on the test-manager:
```
[ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.3.0:wget (download-faban-client) on project benchflow-test-manager: IO Error: Could not get content -> [Help 1]
```

VincenzoFerme:
Related to https://github.com/benchflow/benchflow/issues/274

jeppe-style:
We should include dependencies from BenchFlow in this way:

```
...
    <properties>
        ...
        <main.basedir>${project.basedir}</main.basedir>
        ...
        <benchflow-dsl.project.name>benchflow-dsl</benchflow-dsl.project.name>
        <benchflow-dsl-jar.version>feature-dsl-demo-conversion_046e8d</benchflow-dsl-jar.version>
        <benchflow-dsl.version>0.1.0</benchflow-dsl.version>
        ...
        <benchflow-faban-client.project.name>benchflow-faban-client</benchflow-faban-client.project.name>
        <benchflow-faban-client-jar.version>devel_ad148a</benchflow-faban-client-jar.version>
        <benchflow-faban-client.version>0.1.0</benchflow-faban-client.version>
        ...
        <download-maven-plugin.version>1.3.0</download-maven-plugin.version>
        <maven-install-plugin.version>2.5.2</maven-install-plugin.version>
        ...
    </properties>

    <dependencyManagement>
        <dependencies>
            ...
            <dependency>
                <groupId>cloud.benchflow</groupId>
                <artifactId>benchflow-dsl</artifactId>
[...]

PR #291, opened 2017-05-05
Improves experiment lifecycle

### LINKED ISSUE #223: Maven pom improvements
_From @jeppe-style on December 17, 2016 10:9_

- [ ] groupId should probably be set to `cloud.benchflow.experiment-manager`
- [ ] version and groupId can be inherited from parent pom
- [ ] check that all versions are set by properties

_Copied from original issue: benchflow/experiments-manager#49_

### LINKED ISSUE #254: Integration tests with Docker Compose does not run with maven properly
When running integration tests with docker compose integration using `mvn verify` there are in some (random) tests `java.net.ConnectException: Connection refused (Connection refused)` errors, which makes the test fail.

When the same tests are run with the wercker pipeline (e.g. `wercker dev --pipeline=build-benchflow-experiment-manager`) it is fine. Also running individual tests seems to work fine.

jeppe-style:
In addition to improving the experiment lifecycle:
- fixes #254 
- fixes #223 
- adds style-checker to experiment-manager
- fix for issue (test-manager) https://github.com/benchflow/benchflow/issues/279


VincenzoFerme (commented):
I trust you on the changes. They are all in the same commit, with style changes as well so it is hard to review. I guess you mirrored the changed applied together on the test manager also here, so it is fine. If you want to point me to specific classes ping me and I'll review them.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/models/BenchFlowExperimentModel.java
```diff
@@ -27,6 +27,9 @@
   private BenchFlowExperimentState state;
   private RunningState runningState;
   private TerminatedState terminatedState;
+  private int numTrials;
+  // TODO - this should be part of the DSL
```
VincenzoFerme: I think we agreed that the number of retry of the interactions with Faban is something related to internal configuration only. Or probably not, because this is related to the RetryPolicyTask in case something unexpected happen during the execution, right?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/start/StartTask.java
```diff
@@ -77,8 +72,8 @@ public void run() {
 
       int nTrials = experiment.configuration().terminationCriteria().get().experiment().number();
 
-      // save experiment model in DB
-      experimentModelDAO.addExperiment(experimentID);
+      // save number of trials
+      experimentModelDAO.setNumTrials(experimentID, nTrials);
```
VincenzoFerme: This assumes you always know the number of trials beforehand, that might not be the case.

VincenzoFerme:
What are the assumptions you actually have in this code, that are specific to the supported case, but that cannot be generalised to the cases we decided to enable. See for example: https://github.com/benchflow/benchflow/pull/291#discussion_r115153106

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/start/StartTask.java
jeppe-style: yes - currently the implementation only supports fixed number of trials

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/models/BenchFlowExperimentModel.java
jeppe-style: yes - you are referring to the number of times we should retry to submit the trial to Faban. This number is the number of times we want to run an experiment again if there is some random error during the execution.

jeppe-style:
limitations:
- fixed number of trials
- no internal control of which experiment is executed (need `TaskScheduler`)
- polling of Faban since there is no faban-manager yet
- converts to previous DSL version for compatibility with driver-maker
- limited handling of fatal failure (e.g. only trial had failure more than # trial retries)

VincenzoFerme:
Describe the assumption in an issue dedicated to solve them. A second issue for the test manager if also there you have similar assumptions. - Issue: https://github.com/benchflow/benchflow/issues/332

VincenzoFerme (approved):

Issue #290, opened 2017-05-05
Use natural numbers (>= 1) for test, experiment, trial numbers


jeppe-style:
To make the numbers easier to understand for the users use number 1...n (not 0).

Use the number as index in arrays.

PR #334, opened 2017-05-11
Use natural numbers for test, experiment and trial IDs

jeppe-style:
Fixes https://github.com/benchflow/benchflow/issues/290


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/ExperimentTaskController.java
```diff
@@ -185,34 +188,39 @@ private void handleReExecuteTrial(String experimentID) {
 
     logger.info("handleReExecuteTrial: " + experimentID);
 
-    ReExecuteTrialTask reExecuteTrialTask = new ReExecuteTrialTask(experimentID);
+    try {
 
-    Future<RunStatus> future = experimentTaskExecutorService.submit(reExecuteTrialTask);
+      String trialID = experimentModelDAO.getLastExecutedTrialID(experimentID);
```
VincenzoFerme, line 193: You should know the trial when you access this method, without the need of querying the model.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/models/BenchFlowTestNumber.java
```diff
@@ -14,7 +14,7 @@
 
   @Id private String testIdentifier;
 
-  private Long counter = 0L;
+  private Long counter = 1L;
```
VincenzoFerme, line 17: How do you use this number to increment the test number if the user has already executed a test with the same name?

VincenzoFerme (changes requested):

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/models/BenchFlowTestNumber.java
jeppe-style: In the constructor you can see that it takes a a `userName` and `testName` as input meaning that there will be one `BenchFlowTestNumber` object stored in the DB per combination. In the [TestModelDAO](https://github.com/benchflow/benchflow/blob/408297ca24ea6f12d257f9c4ca33b55ea3e236ef/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/services/internal/dao/BenchFlowTestModelDAO.java#L43) class you can see how the test number is generated.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/ExperimentTaskController.java
jeppe-style: yes we need to rework this according to the updates we made to the lifecycle.

VincenzoFerme: Can you link the issue?

jeppe-style: here is the issue https://github.com/benchflow/benchflow/issues/347

VincenzoFerme (approved):

Issue #283, opened 2017-05-05
CodeChecker warnings that could be fixed automatically


jeppe-style:
When running the code checker in Java some changes could be automatically fixed:
- `SingleLineJavadoc Single-line Javadoc comment should be multi-line.`
- `NeedBraces	'if' construct must use '{}'s.`
- `CustomImportOrder	Wrong lexicographical order for 'javax.validation.Valid' import. Should be before 'org.slf4j.LoggerFactory'.`

igormoreno:
I am aware of the 1st case. I'm working on it. The other ones I have never seen. Where are you getting those? Is it in the test-manager? Could you tell me the hashes of the specific commits so I can reproduce here? Thank you :-)

jeppe-style:
@igormoreno ok - try with this commit https://github.com/benchflow/benchflow/pull/266/commits/b4f8adb9ac869c47bd799d03d971161bbd41fc31

igormoreno:
I'm getting:
```
[ERROR] Failed to execute goal on project application: Could not resolve dependencies for project cloud.benchflow:application:jar:0.1.0: The following artifacts could not be resolved: cloud.benchflow:benchflow-dsl:jar:0.1.0, cloud.benchflow.faban:client:jar:0.1.0: Could not find artifact cloud.benchflow:benchflow-dsl:jar:0.1.0 in palantir (https://dl.bintray.com/palantir/releases) -> [Help 1]
```
on this commit

igormoreno:
@jeppe-style  I'm getting:
```
[ERROR] Failed to execute goal on project application: Could not resolve dependencies for project cloud.benchflow:application:jar:0.1.0: The following artifacts could not be resolved: cloud.benchflow:benchflow-dsl:jar:0.1.0, cloud.benchflow.faban:client:jar:0.1.0: Could not find artifact cloud.benchflow:benchflow-dsl:jar:0.1.0 in palantir (https://dl.bintray.com/palantir/releases) -> [Help 1]
```
on this commit

igormoreno:
The solution for these problems is the following:
- `SingleLineJavadoc Single-line Javadoc comment should be multi-line.`
[...]

PR #336, opened 2017-05-12
Various improvements to devops static checks

### LINKED ISSUE #329: Scalastyle Checker Runs also for non Scala Projects
When I build, e.g., the `test-manager` I also see scalastyle related logs. This should not happen because the `test-manager` does not contain scala code.

An example of the logs:

```
[WARNING] sourceDirectory is not specified or does not exist value=/<my_local_folder>/benchflow-test-manager/src/main/scala
[WARNING] testSourceDirectory is not specified or does not exist value=/<my_local_folder>/benchflow-test-manager/src/test/scala
Saving to outputFile=/<my_local_folder>/scalastyle-output.xml
Processed 0 file(s)
Found 0 errors
Found 0 warnings
Found 0 infos
Finished in 115 ms
```

igormoreno:
* Setting up PMD (which already runs in continuous integration)
* Run scalacheck only on scala projects (closes #329)
* Activate static checks in all remaining java projects
* Changing Java formatting plug-in (#283)

### DIFF
--- a/benchflow-driver-maker/pom.xml
+++ b/benchflow-driver-maker/pom.xml
@@ -1,6 +1,13 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
+
+    <parent>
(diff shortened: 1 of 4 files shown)


VincenzoFerme:
@igormoreno Why do we need eclipse-java-google-style.xml?

igormoreno:
The new formatter uses a configuration file describing how the code should be formatted.
http://code.revelc.net/formatter-maven-plugin/formatter-maven-plugin/index.html

The configuration file was taken from here:
https://github.com/google/styleguide

VincenzoFerme (approved):

PR #354, opened 2017-05-14
Fix various static check issues on experiment manager

igormoreno:
Issues reported in #283


VincenzoFerme:
Does this pull request fixes completely https://github.com/benchflow/benchflow/issues/283? If yes, move to documentation what has been discussed in https://github.com/benchflow/benchflow/issues/283, and mark this pull request as *fixes* the https://github.com/benchflow/benchflow/issues/283.

igormoreno:
I wanted to use the "fixes" keyword like I did in other commits but unfortunately in this case it doesn't work because I decided to create different pull requests for changes in different projects. So I will close the issue manually in the end.

VincenzoFerme (approved):

PR #355, opened 2017-05-15
Fix various static check issues in faban client

igormoreno:
The remaining issues in the faban client are all caused by lack of proper Javadoc documentation. I think it's probably more productive for someone that already has knowledge of the code base to document these methods.

I want to add that unlike traditional Javadoc checks, the current checks are very reasonable so it's worth to follow them. Here's part of the set up right now (which we can also find to if we decide to):
```
        <module name="JavadocMethod">
            <property name="scope" value="public"/>
            <property name="allowMissingParamTags" value="true"/>
            <property name="allowMissingThrowsTags" value="true"/>
            <property name="allowMissingReturnTag" value="true"/>
            <property name="minLineCount" value="2"/>
            ...
        </module>
```

This means that:
- only public methods have to be documented
- it is not mandatory to document each single parameter, thrown exception, or return. This is good because in practice sometimes only a textual description is enough. The traditional checks force you to document each single parameter which leads to useless documentation like "experimentID: the ID of the experiment". But it is important to notice that if you do write an "@param someParameter" or a "@throws SomeException" without documentation the static check will complain
- methods with less than 3 lines (we can change this parameter if you want) don't have to be documented. That is also great because if you're forced to document obvious code like getters and setters then nobody takes the static checks seriously

### DIFF
--- a/benchflow-faban-client/src/main/java/cloud/benchflow/faban/client/FabanClient.java
+++ b/benchflow-faban-client/src/main/java/cloud/benchflow/faban/client/FabanClient.java
@@ -1,11 +1,34 @@
 package cloud.benchflow.faban.client;
 
-import cloud.benchflow.faban.client.commands.*;
-import cloud.benchflow.faban.client.configurations.*;
-import cloud.benchflow.faban.client.exceptions.*;
-import cloud.benchflow.faban.client.responses.*;
+import cloud.benchflow.faban.client.commands.DeployCommand;
+import cloud.benchflow.faban.client.commands.KillCommand;
+import cloud.benchflow.faban.client.commands.PendingCommand;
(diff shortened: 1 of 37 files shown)


VincenzoFerme:
@igormoreno i fixed the warnings related to https://github.com/benchflow/benchflow/pull/355#issue-228661949. I approved the pull request because for me it is ready, just check that what I did is sufficient, and move the information present in https://github.com/benchflow/benchflow/pull/355#issue-228661949 in documentation, and I can proceed to merge this pull.

VincenzoFerme (dismissed):

igormoreno:
I read the documentation you added. It's very nice :-)
I just pushed a commit updating the documentation as you requested.

Issue #319, opened 2017-05-06
Refactor TaskController to TaskScheduler and Add Running and Ready Queues


jeppe-style:
In the experiment-manager and test-manager we want to add two queues for test/experiments:
- _ready queue_: when an test/experiment has completed the start state it will be put here
- _running queue_: a queue that contains (for now one) the currently running test/experiment

In addition the `TaskScheduler` will have one `ExecutorService` that executes the different tasks according to available resources.

We have these two queues to control how many test/experiments can run at the same time. They can be `BlockingQueue`s (see [JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html)). The two queues are FIFO.

The general idea is to keep all the business logic/control flow in the `TaskScheduler`.

As a request comes in a start task is scheduled on the executor service. When finished the test/experiment ID is put in the _ready queue_ and its state changed to `READY`.

Once a test/experiment is `READY` it is checked if the _running queue_ is free. If it is, the state is changed to `RUNNING` and the test/experiment is executed until it has reached state `TERMINATED` or `WAITING` (only test manager - when user input is needed). Otherwise nothing happens.

When a test/experiment is `paused` the state changes to `READY`, however, it is not put in the _ready queue_. Once the input is given to continue execution it is added to the _ready queue_ (by calling `handleExperimentState(...)`).

Once the test/experiment has reached the state `TERMINATED`, or `WAITING` it is removed from the _running queue_ and the next test/experiment in the _ready queue_ is processed.

VincenzoFerme:
@jeppe-style in the description above you mention the `WAITING` state, but this state is only related to the `test-manager` in https://github.com/VincenzoFerme/benchflow-temporary-documentation/blob/master/roadmap/Jesper/architecture/lifecycle-sequence.pdf. What is the equivalent in the `experiment-manager` for scheduling the next trial? Is it the `ABORTED` state?

jeppe-style:
Yes - you are right it, is not clear. I updated the text above with some more clarification also about `pausing`. There is no `WAITING` state for the experiment.

VincenzoFerme:
You change make me thing that we need a `PAUSED` state.

jeppe-style:
When would that be the case?

VincenzoFerme:
When it is `PAUSED` because, they now switch to `READY` but they do not go in the ready queue. To me something in a given state, has always to have the same behaviour and the same things have to happen.

[...]

PR #356, opened 2017-05-21
Adds Task Scheduler

jeppe-style:
Fixes issue for experiment-manager https://github.com/benchflow/benchflow/issues/319


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentDispatcher.java
```diff
@@ -0,0 +1,79 @@
...
+  public void run() {
+
+    logger.info("running");
+
+    while (true) {
+
+      try {
+
+        // wait for next experiment to be ready
+        String experimentID = readyQueue.take();
```
VincenzoFerme, line 45: ``The call to queue.take() blocks the thread that invoked it until something is available in the queue. This means that the thread will wait there endlessly, unless interrupted, until an item is added to the queue.`` Is this something wanted right? So this way you put thread to the waiting state, before the next iteration of the `while` to happen?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -15,43 +15,50 @@
...
+  public ExperimentTaskScheduler(ExecutorService experimentTaskExecutorService) {
 
     this.experimentTaskExecutorService = experimentTaskExecutorService;
 
     this.experimentModelDAO = BenchFlowExperimentManagerApplication.getExperimentModelDAO();
     this.trialModelDAO = BenchFlowExperimentManagerApplication.getTrialModelDAO();
     this.testManagerService = BenchFlowExperimentManagerApplication.getTestManagerService();
+
+    // start dispatcher in separate thread
+    new Thread(new ExperimentDispatcher(readyQueue, runningQueue)).start();
```
VincenzoFerme, line 60: Is there a better way to start this independently of the constructor of the `ExperimentTaskScheduler`? I understand you have to pass the queues, but it looks weird to me that the Scheduler starts the Dispatcher.

VincenzoFerme (changes requested):

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentDispatcher.java
jeppe-style: The experiment dispatcher runs in a separate thread so it should not block anything else than that thread (e.g. itself).
Basically the call is there to ensure that some experiment is in the ready queue waiting to be scheduled.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: Yes, I agree that it looks weird and it is not an optimal solution. The two options I can think of are:
- start it in a separate method called `initialize`
- start it directly in the application right after the `ExperimentTaskScheduler` has been initialized - although then we have to somehow pass the queues

VincenzoFerme: Or share the queues with the package modifier, so that all the classes in the `scheduler` package can access the queues (that I think it should be because we should only have classes needing the queues in the `scheduler` package). Then it it fine that they are initialised by the main actor in the package, that is the `ExperimentTaskScheduler`, that is always the first class that get instantiated in the main application. This way you can initialise also the dispatcher in the main application class.

jeppe-style: I'm not so in favor of sharing data structures in that way since it creates strong coupling. Also we have to depend on the order of how objects are instantiated. I know I have that someplace else, but I would avoid it if possible. I would then prefer to call `initialize` from the main application.
In fact maybe all classes (except resources) that get instances from the main application should have an `initialize` method where these instances are assigned. Then it would be two clear steps in the main application:
- instantiate the objects
- call initialize on each instance to assign dependencies

VincenzoFerme: I agree with you, so proceed as you suggested and fix also the "someplace else" where you have the same ;).

jeppe-style: I added a new issue for this so I can do it in a separate pull-request https://github.com/benchflow/benchflow/issues/359

VincenzoFerme (approved):

Issue #341, opened 2017-05-13
Add Generation of Docker Compose from Parameters


jeppe-style:
as title

support for:
- memory
- CPU
- environment variables

VincenzoFerme:
@jeppe-style is this closed by https://github.com/benchflow/benchflow/pull/368 or does it still need improvements?

jeppe-style:
@VincenzoFerme that is fine - there is the issue https://github.com/benchflow/benchflow/issues/380 that covers missing parts

PR #368, opened 2017-05-25
DSL docker-compose generation

jeppe-style:
Adds support for generating docker-compose with memory limit and environment variables. 

In reference to https://github.com/benchflow/benchflow/issues/341

### DIFF
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/dockercompose/DockerComposeYamlBuilder.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/dockercompose/DockerComposeYamlBuilder.scala
@@ -1,0 +1,67 @@
+package cloud.benchflow.dsl.dockercompose
+
+import cloud.benchflow.dsl.definition.types.bytes.Bytes
+import cloud.benchflow.dsl.definition.types.bytes.BytesYamlProtocol._
+import net.jcazevedo.moultingyaml.{ YamlString, YamlValue, _ }
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com)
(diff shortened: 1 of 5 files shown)


benchflow-dsl/src/main/scala/cloud/benchflow/dsl/dockercompose/DockerComposeYamlBuilder.scala
```diff
@@ -0,0 +1,67 @@
...
+import net.jcazevedo.moultingyaml.{ YamlString, YamlValue, _ }
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com)
+ *         created on 2017-05-25
+ */
+class DockerComposeYamlBuilder(dockerComposeYamlValue: YamlValue) {
+
+  val ServicesKey = YamlString("services")
+  val MemLimitKey = YamlString("mem_limit")
```
VincenzoFerme, line 14: Is this the hard or the soft limit? Is it ok to support only the hard limit, but we should name the variables in a consistent way so that then we can extend the DSL.

jeppe-style: This is what I could find for docker-compose `v2` (https://docs.docker.com/compose/compose-file/compose-file-v2/#restart). 
I think it would be good to discuss  which version we want to support.

VincenzoFerme: We discussed and the outcome is issue #380

VincenzoFerme (approved):

Issue #342, opened 2017-05-13
Add Generation of Experiment from Configuration Parameters


jeppe-style:
as per title

jeppe-style:
This has actually already been implemented since it is currently only needed for number of users.

@VincenzoFerme the issue can be closed.

VincenzoFerme:
Is it implemented for all the all the configuration parameters we decided to support?

jeppe-style:
Yes, cpu, memory and environment variables only influences the docker-compose file so those are handled separately in https://github.com/benchflow/benchflow/issues/341

Issue #282, opened 2017-05-04
Support Different Configuration Settings Exploration


VincenzoFerme:
Support also other settings for the exploration, other than the users.

Issue #370, opened 2017-05-27
Generate and store exploration space


jeppe-style:
Precompute exploration spaces for `one-at-a-time` and `random-breakdown` and make them available through API call

This has been changed (see below)

VincenzoFerme:
Why not also for the other strategies?

jeppe-style:
Because they cannot be pre-computed

VincenzoFerme:
Why not, the full space should always be possible to be precomputed.

jeppe-style:
but only in `one-at-a-time` or `random` order

I'm thinking now that also random order we actually don't need to precompute, but just keep a list with the order of experiments (e.g. from 0 to exploration space size) which we can shuffle to give us randomness. This is easy to do in the test-manager.

VincenzoFerme:
Probably I miss what you mean with `exploration-space` then. Can you explain me what are you precomputing here, for which strategies you can precompute and for which no and why? Thanks :)

jeppe-style:
Yes! Sorry I'm very much into it so it is difficult to step out.

In general I define an instance that summarizes the exploration space defined by the user in the test definition. Here each dimension (e.g. memory, number of users) is stored as lists with all possible values for that dimension.

From that I define an instance to keep the state of the exploration space, e.g. with capacity to store all possible combinations of dimensions. In this instance I keep a list for each dimension that has the size of all possible combinations. If we have a `one-at-a-time` selection strategy I can precompute the values, e.g. index 0 in the list corresponds to experiment 1, index 1 in the list corresponds to experiment etc. This would is also possible for `random` selection strategy by shuffling the indices, however, we might just want to keep a separate list with the order which we then can shuffle.
For other strategies I insert `-1` into the given indices to indicate that this has not yet been executed, since for other exploration strategies we want to decide the configuration of the next experiment based on the previous results. Here we could also used the precomputed order, however, we then would need to search through to find the particular configuration we want to execute next.

It is a bit difficult to explain this in written, so maybe it is easier to discuss when we meet.

jeppe-style:
Outcome of discussion:
**DSL**
- provides an API to generate the complete exploration space (e.g. the matrix we talked about), where a row corresponds to a configuration of an experiment
- provides an API to get the next experiment bundle (definition, docker compose) based on an exploration space, index to the row, test definition and docker compose
- provides an API to get the next experiment bundle and index to the row (definition, docker compose) based on an exploration space, an experiment configuration, test definition and docker compose

[...]

Issue #257, opened 2017-04-25
Add a layer to decide the next experiment


jeppe-style:
Add a layer that returns the next experiment instead of doing it in the task (has business logic) directly.

VincenzoFerme:
We decided that the `test-manager` decides the next experiment and asks the `dsl` a specific point in the exploration space.

jeppe-style:
@VincenzoFerme I think we can close the issue since both PR have been merged

PR #371, opened 2017-05-27
Adds Generation of the Exploration Space and Experiment Bundle

jeppe-style:
Fixes issue related to the DSL https://github.com/benchflow/benchflow/issues/370

Fixes https://github.com/benchflow/benchflow/issues/342

Fixes https://github.com/benchflow/benchflow/issues/282

Fixes issue related to the DSL https://github.com/benchflow/benchflow/issues/257

### DIFF
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
@@ -1,10 +1,9 @@
 package cloud.benchflow.dsl
 
 import cloud.benchflow.dsl.BenchFlowDSL.testFromYaml
-import cloud.benchflow.dsl.definition.configuration.BenchFlowExperimentConfiguration
 import cloud.benchflow.dsl.definition.errorhandling.BenchFlowDeserializationException
 import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator
-import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator.ExplorationSpaceDimensions
+import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator.{ ExplorationSpace, ExplorationSpaceDimensions, ExplorationSpacePoint }
 
 /**
  * @author Jesper Findahl (jesper.findahl@gmail.com)
@@ -14,7 +13,7 @@ object ExplorationSpace {
 
   type ExperimentDefinitionYamlString = String
   type DockerComposeYamlString = String
-  type ExperimentNumber = Int
+  type ExperimentIndex = Int
 
   /**
    *
@@ -52,46 +51,83 @@ object ExplorationSpace {
   /**
    *
    * @param explorationSpace         the exploration space where the experiment is
-   * @param experimentNumber         the number of the experiment (index)
+   * @param experimentIndex          the index of the experiment
    * @param testDefinitionYamlString the test definition to build the experiment definition from as a YAML string
    * @param dockerComposeYamlString  the docker compose to build the experiment docker-compose from as a YAML string
    * @return a tuple of experiment definition YAML string and docker compose YAML string
    */
   def generateExperimentBundle(
-    explorationSpace: ExplorationSpaceDimensions,
-    experimentNumber: ExperimentNumber,
+    explorationSpace: ExplorationSpace,
+    experimentIndex: ExperimentIndex,
     testDefinitionYamlString: String,
     dockerComposeYamlString: String): (ExperimentDefinitionYamlString, DockerComposeYamlString) = {
 
-    // get the experiment configuration from the exploration space
+    // build the experiment definition
+    var experimentBuilder = BenchFlowDSL.experimentYamlBuilderFromTestYaml(testDefinitionYamlString)
 
-    // generate the experiment definition YAML
+    experimentBuilder = explorationSpace.usersDimension match {
+      case Some(list) => experimentBuilder.numUsers(list(experimentIndex))
+      case None => experimentBuilder
+    }
 
-    // generate the docker compose YAML
-    ("to do", "to do")
+    val experimentDefinitionYamlString = experimentBuilder.build()
+
+    // build the docker compose YAML
+    var dockerComposeBuilder = BenchFlowDSL.dockerComposeYamlBuilderFromDockerComposeYaml(dockerComposeYamlString)
+
+    explorationSpace.memoryDimension match {
+      case Some(map) => map foreach {
+        case (serviceName, list) => dockerComposeBuilder = dockerComposeBuilder.memLimit(serviceName, list(experimentIndex))
+      }
+      case None => // nothing to do
+    }
+
+    explorationSpace.environmentDimension match {
+      case Some(serviceMap) => serviceMap foreach {
+        case (serviceName, environmentMap) => environmentMap foreach {
+          case (environmentVariable, list) => dockerComposeBuilder = dockerComposeBuilder.environmentVariable(
+            serviceName,
+            environmentVariable,
+            list(experimentIndex))
+        }
+      }
+      case None => // nothing to do
+    }
+
+    val newDockerComposeYamlString = dockerComposeBuilder.build()
+
+    (experimentDefinitionYamlString, newDockerComposeYamlString)
   }
 
   /**
+   * Generates and experiment bundle from the given exploration space point. Assumes the point exists in the
+   * exploraiton space.
    *
    * @param explorationSpace         the exploration space where the experiment is
-   * @param experimentConfiguration  the configuration for the experiment to be generated
+   * @param explorationSpacePoint    the point in the exploration space from which to generate the experiment
    * @param testDefinitionYamlString the test definition to build the experiment definition from as a YAML string
    * @param dockerComposeYamlString  the docker compose to build the experiment docker-compose from as a YAML string
-   * @return a tuple of experiment definition YAML string, docker compose YAML string and experiment number (index in exploration space)
+   * @return a tuple of experiment definition YAML string, docker compose YAML string and experiment number
+   *         (index in exploration space) as an Option. None if point could not be found.
    */
   def generateExperimentBundle(
-    explorationSpace: ExplorationSpaceDimensions,
-    experimentConfiguration: BenchFlowExperimentConfiguration,
+    explorationSpace: ExplorationSpace,
+    explorationSpacePoint: ExplorationSpacePoint,
     testDefinitionYamlString: String,
-    dockerComposeYamlString: String): (ExperimentDefinitionYamlString, DockerComposeYamlString, ExperimentNumber) = {
+    dockerComposeYamlString: String): Option[(ExperimentDefinitionYamlString, DockerComposeYamlString, ExperimentIndex)] = {
 
-    // find the index of the given experiment configuration in the exploration space
+    ExplorationSpaceGenerator.getExperimentIndex(explorationSpace, explorationSpacePoint) match {
+      case None => None
+      case Some(experimentIndex) =>
 
-    // generate the experiment definition YAML
+        val (experimentDefinitionYamlString, newDockerComposeYamlString) = generateExperimentBundle(
+          explorationSpace,
+          experimentIndex,
+          testDefinitionYamlString,
+          dockerComposeYamlString)
 
-    // generate the docker compose YAML
-
-    ("to do", "to do", -1)
+        Some(experimentDefinitionYamlString, newDockerComposeYamlString, experimentIndex)
+    }
 
   }
 
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
@@ -1,15 +1,20 @@
 package cloud.benchflow.dsl
 
 import cloud.benchflow.dsl.BenchFlowDSL.testFromYaml
+import cloud.benchflow.dsl.definition.configuration.BenchFlowExperimentConfiguration
 import cloud.benchflow.dsl.definition.errorhandling.BenchFlowDeserializationException
 import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator
-import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator.{ ExplorationSpace, ExplorationSpaceState }
+import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator.ExplorationSpaceDimensions
 
 /**
  * @author Jesper Findahl (jesper.findahl@gmail.com)
  *         created on 2017-06-01
  */
 object ExplorationSpace {
+
+  type ExperimentDefinitionYamlString = String
+  type DockerComposeYamlString = String
+  type ExperimentNumber = Int
 
   /**
    *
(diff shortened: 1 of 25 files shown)


VincenzoFerme:
@jeppe-style Lets finalise the leftover work to improve what we discussed in this pull request.

jeppe-style:
@VincenzoFerme ok - I'll add the comments also here then from https://github.com/benchflow/benchflow/pull/369#issuecomment-304846998

VincenzoFerme:
Reference the issue that discuss how this should be done.

jeppe-style:
It is the issue already referenced.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -0,0 +1,98 @@
...
+import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator
+import cloud.benchflow.dsl.explorationspace.ExplorationSpaceGenerator.ExplorationSpaceDimensions
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com)
+ *         created on 2017-06-01
+ */
+object ExplorationSpace {
+
+  type ExperimentDefinitionYamlString = String
```
VincenzoFerme, line 14: Do you need the entire definition? Isn't it enough to have the goal definition part?

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -0,0 +1,98 @@
...
+    experimentNumber: ExperimentNumber,
+    testDefinitionYamlString: String,
+    dockerComposeYamlString: String): (ExperimentDefinitionYamlString, DockerComposeYamlString) = {
+
+    // get the experiment configuration from the exploration space
+
+    // generate the experiment definition YAML
+
+    // generate the docker compose YAML
+    ("to do", "to do")
```
VincenzoFerme: In this or another pull? If another pull, reference an issue here.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -0,0 +1,98 @@
...
+    testDefinitionYamlString: String,
+    dockerComposeYamlString: String): (ExperimentDefinitionYamlString, DockerComposeYamlString, ExperimentNumber) = {
+
+    // find the index of the given experiment configuration in the exploration space
+
+    // generate the experiment definition YAML
+
+    // generate the docker compose YAML
+
+    ("to do", "to do", -1)
```
VincenzoFerme: In this or another pull? If another pull, reference an issue here.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
```diff
@@ -63,112 +75,114 @@ object ExplorationSpaceGenerator {
...
 
       }
 
     }
 
   }
 
-  def fillList(shift: Int, numValues: Int, explorationSpaceSize: Int): List[Index] = {
-
+  /**
```
VincenzoFerme: Why do you need this function?

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
```diff
@@ -11,30 +11,41 @@ import scala.collection.mutable
...
   type VariableName = String
   type VariableValue = String
   type NumUsers = Int
   type DimensionLength = Int
   type Index = Int
 
-  case class ExplorationSpace(
+  case class ExplorationSpaceDimensions(
     users: Option[List[NumUsers]],
     memory: Option[Map[ServiceName, List[Bytes]]],
```
VincenzoFerme: This is actually a `resource` in the DSL. I think would be better to keep the same nesting also in the code.

VincenzoFerme: The same for other cases

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -52,46 +51,83 @@ object ExplorationSpace {
...
-
-    // generate the experiment definition YAML
-
-    // generate the docker compose YAML
-    ("to do", "to do")
+    // build the experiment definition
+    var experimentBuilder = BenchFlowDSL.experimentYamlBuilderFromTestYaml(testDefinitionYamlString)
+
+    experimentBuilder = explorationSpace.usersDimension match {
+      case Some(list) => experimentBuilder.numUsers(list(experimentIndex))
```
VincenzoFerme, line 69: What happens when you do not know the experiment index?

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -52,46 +51,83 @@ object ExplorationSpace {
...
+    }
+
+    val newDockerComposeYamlString = dockerComposeBuilder.build()
+
+    (experimentDefinitionYamlString, newDockerComposeYamlString)
   }
 
   /**
+   * Generates and experiment bundle from the given exploration space point. Assumes the point exists in the
+   * exploraiton space.
```
VincenzoFerme: Typo

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -52,46 +51,83 @@ object ExplorationSpace {
...
+      case None => // nothing to do
+    }
+
+    val newDockerComposeYamlString = dockerComposeBuilder.build()
+
+    (experimentDefinitionYamlString, newDockerComposeYamlString)
   }
 
   /**
+   * Generates and experiment bundle from the given exploration space point. Assumes the point exists in the
```
VincenzoFerme: You should check it exists, not assume it. Even though it is you system calling this method.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
```diff
@@ -194,4 +201,75 @@ object ExplorationSpaceGenerator {
     explorationSpaceSizeOption
 
   }
+
+  /**
```
VincenzoFerme, line 208: With a second data structure like the following, the match can be constant in time using the columns:

```
           user:5, user:10, service_a:memory:1GB, .....
row1
row2
row3
...
```

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
```diff
@@ -194,4 +201,75 @@ object ExplorationSpaceGenerator {
...
+   * Find the index of the given experiment configuration in the exploration space
+   * by traversing the space and performing the intersection of the found possible indices
+   *
+   * @param explorationSpace
+   * @param explorationSpacePoint
+   * @return
+   */
+  def getExperimentIndex(
```
VincenzoFerme, line 216: Add comments, there I can review it

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/ExplorationSpaceTest.scala
```diff
@@ -0,0 +1,104 @@
...
+
+    Assert.assertTrue(experimentDefinition.contains("users: " + 5))
+
+    Assert.assertTrue(newDockerCompose.contains("SIZE_OF_THREADPOOL=3"))
+    Assert.assertTrue(newDockerCompose.contains("AN_ENUM=C"))
+    Assert.assertTrue(newDockerCompose.contains("mem_limit: 1g"))
+
+  }
+
+  @Test def generateExperimentBundleWithConfigurationTest(): Unit = {
```
VincenzoFerme, line 69: Where is a test testing all the expected generated experiments?

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
jeppe-style: I don't get the question fully

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
jeppe-style: I wanted remove the nesting so it is more clear and to avoid having to access another layer

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
jeppe-style: This is a return type

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
VincenzoFerme: It is fine, but please add some comments

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
VincenzoFerme: There is another function handling it

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/ExplorationSpaceTest.scala
VincenzoFerme: This is about an experiment bundle.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
jeppe-style: I just saw this was an old commit. It actually returns an option, e.g., `None` if the point doesn't exist (or more than one). I will update the JavaDoc.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
jeppe-style: Here is an issue for reference https://github.com/benchflow/benchflow/issues/386

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/ExplorationSpaceTest.scala
jeppe-style: it is in another file https://github.com/benchflow/benchflow/pull/371/files/7ae4fe21d3437049d02e21af0f53c94e7da611f0..402fc9aee93e331b276badfd9f39a92cdee88291#diff-173d315759c44281b5ce418acd20b7fcR74

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
jeppe-style: comments have been added

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
```diff
@@ -100,15 +100,14 @@ object ExplorationSpace {
...
-   * exploraiton space.
+   * Generates and experiment bundle from the given exploration space point.
    *
    * @param explorationSpace         the exploration space where the experiment is
    * @param explorationSpacePoint    the point in the exploration space from which to generate the experiment
    * @param testDefinitionYamlString the test definition to build the experiment definition from as a YAML string
    * @param dockerComposeYamlString  the docker compose to build the experiment docker-compose from as a YAML string
    * @return a tuple of experiment definition YAML string, docker compose YAML string and experiment number
-   *         (index in exploration space) as an Option. None if point could not be found.
+   *         (index in exploration space) as an Option. None if point could not be found, or if multiple were found.
```
VincenzoFerme: You need two different return types for the described cases. If multiple have been found it has to throw an error.

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
jeppe-style: I use an [Option](https://www.scala-lang.org/api/current/scala/Option.html) for that. The calling class has to unwrap that and handle the two cases (`Some` or `None`).

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
jeppe-style: Isn't it clear from the description?
It is used to create the combinations of a given dimension (each dimension is stored in a list) to produce the cartesian product.

VincenzoFerme: It is not, add a comment with an example that clarify what is the input and what you actually do in this method, please.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
VincenzoFerme: @jeppe-style I understand but how do you handle the difference between 0 point and multiple points? I don’t see any difference. I think it is enough to say that this is not a valid configuration.

VincenzoFerme: Ah, now it is clear, make it explicit in the Javadoc please.

VincenzoFerme: The important thing to me is that in the caller you can distinguish the two cases, because you need for sure to generate different error messages.

jeppe-style: I updated the JavaDoc to clarify when `Some` or `None` is returned.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/explorationspace/ExplorationSpaceGenerator.scala
jeppe-style: The examples are already there, is this really not enough?
```
Fills a list with indices for calculating the cartesian product.
If we want to create the list [0,0,1,1,2,2] we would pass: blockSize = 2, numValues = 3, listLength = 6
If we want to create the list [0,1,2,0,1,2] we would pass: blockSize = 1, numValues = 3, listLength = 6
```

VincenzoFerme (approved):

Issue #384, opened 2017-05-31
Change scalastyle rules when running on tests


igormoreno:
I think we shouldn't run scalacheck on tests or at least change the rules for it because it's generating issues that don't really make sense for tests like "magic number" complaints, for example.

PR #390, opened 2017-06-04
Adds Java compatible exploration space types

jeppe-style:
To be able to store the exploration space data structures in the DB on the test-manager we need to change the `Bytes` type to a string because it contains a Scala enumeration that is not compatible with Java.

When this pull-request is complete it fixes partially https://github.com/benchflow/benchflow/issues/391 with a solution that add another object that is returned from the API. A better solution is discussed in the referenced issue, that hide the conversions in the scala objects, so that once one get a scala object it can call a method `asJava()` to get the Java representation.

depends on PR https://github.com/benchflow/benchflow/pull/465


VincenzoFerme:
@jeppe-style why do you just not define a type that is compatible so you can avoid conversion?

jeppe-style:
I could not find any good other solution. If you have any suggestion for an alternative solution I'm happy to hear about that.

VincenzoFerme:
@jeppe-style open an issue describing the problem solved by this pull request, and I can give it a try.

jeppe-style:
Thanks! I created an issue here https://github.com/benchflow/benchflow/issues/391

Update: this is no longer necessary since we can use `Bytes` directly

VincenzoFerme:
@jeppe-style what should we do with this pull request?

jeppe-style:
I'll work on that as I go down the issues in the Demo issue

jeppe-style:
The current errors in Codacy should not be reported: https://www.codacy.com/app/BenchFlow/benchflow/pullRequest?prid=689856

VincenzoFerme:
@jeppe-style I disabled the warning for the import line that was too long, but we should fix the magic number issue (even if it test code, is code that has to be maintained): https://www.codacy.com/app/BenchFlow/benchflow/pullRequest?prid=689856

jeppe-style:
@VincenzoFerme I discussed magic numbers also with Igor and he also agreed that this check doesn't make sense for tests. It was discussed in this issue https://github.com/benchflow/benchflow/issues/384.

VincenzoFerme:
I had to resolve conflict for merging this PR with devel. @jeppe-style please check if the version of the following files I kept are the expected ones (I link the version I kept and the PR it belongs to so you can also check when I kept versions from this PR and when from other PRs):

- https://github.com/benchflow/benchflow/pull/390/files#diff-04aed422be36d830a6822d5784e5a436
- https://github.com/benchflow/benchflow/blob/f76d6f2b19bb26e98ca9a2406a487cdb41165eb6/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/ExplorationValuesBytesYamlProtocol.scala
- ExplorationSpace.scala (kept the version from devel)
- https://github.com/benchflow/benchflow/blob/d71996036c9bc60e6ee74d6258300b8aca2cc10a/benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/ByteValuesTest.scala

VincenzoFerme:
@jeppe-style can you please check the code is compliant with the complete discussion we have on https://github.com/benchflow/benchflow/issues/391?

jeppe-style:
@VincenzoFerme yes - this fits and can be saved in the DB

VincenzoFerme (approved):

Issue #385, opened 2017-06-02
Add Possibility to Start Service in Test Mode


jeppe-style:
To be able to perform basic component testing of a service it should be able to start in a test mode to mock external services, e.g. Faban, to test the functionality without actually running an experiment.

This should be done by adding the following in the configuration file:
```
testMode:
  mockFaban: ${env.MOCK_FABAN!false?c}
```

When the service starts in this mode with the boolean flag for a given service set to `true` that service  should be mocked. The mode can also be set as an environment variable `MOCK_FABAN`.  `?c` is need to be able to [convert from a string to boolean](http://freemarker.org/docs/ref_builtins_boolean.html#ref_builtin_c_boolean)

To control how the mock should behave we create scenarios. The mocked service understands the scenario by the name of the experiment, e.g., if the name is `scenario1` then the mock knows it should simulate scenario 1.

This is done to make it easier to control the expected behavior.

VincenzoFerme:
@jeppe-style give a look at the following comment, because can help with this as well: https://github.com/benchflow/benchflow/issues/348#issuecomment-305913034

jeppe-style:
@VincenzoFerme I had a look, however, I don't see how this can help in our case since we want to control it from the configuration and the example is about overriding the configuration for testing purposes and insert mocks that have been constructed in the test classes.

VincenzoFerme:
You need to define a configuration section that works both when the test mode is needed, but also when it is not needed. I mean, the info about the testing (if enabled or not) have to be specified using an environment variable as well.

jeppe-style:
Yes I agree and already had included that in the implementation. I updated the initial note accordingly.

VincenzoFerme:
Closed by https://github.com/benchflow/benchflow/pull/392

PR #392, opened 2017-06-04
Adds test mode to mock Faban with basic scenarios

jeppe-style:
Adds test mode to mock Faban with basic scenarios. The scenario to use is specified by changing the `test name` in the following way for each trial (the case of the letters doesn't matter):
- always successful execution - `alwaysCompleted`,  also default if the `test name` does not match any other scenario
- fail on first execution and then always success - `failFirstExecution`
- always failed execution - `alwaysFail`

Fixes issue in BenchFlowExperimentManagerApplicationIT that did not store the mock.bpmn model on Minio before running tests

Fixes https://github.com/benchflow/benchflow/issues/385

### DIFF
--- a/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplication.java
+++ b/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplication.java
@@ -2,6 +2,7 @@ package cloud.benchflow.experimentmanager;
 
 import cloud.benchflow.experimentmanager.configurations.BenchFlowExperimentManagerConfiguration;
 import cloud.benchflow.experimentmanager.resources.BenchFlowExperimentResource;
+import cloud.benchflow.experimentmanager.scheduler.ExperimentTaskScheduler;
 import cloud.benchflow.experimentmanager.services.external.BenchFlowTestManagerService;
 import cloud.benchflow.experimentmanager.services.external.DriversMakerService;
 import cloud.benchflow.experimentmanager.services.external.FabanManagerService;
@@ -9,11 +10,8 @@ import cloud.benchflow.experimentmanager.services.external.MinioService;
 import cloud.benchflow.experimentmanager.services.external.test.FabanManagerServiceMock;
 import cloud.benchflow.experimentmanager.services.internal.dao.BenchFlowExperimentModelDAO;
 import cloud.benchflow.experimentmanager.services.internal.dao.TrialModelDAO;
-import cloud.benchflow.experimentmanager.scheduler.ExperimentTaskScheduler;
 import cloud.benchflow.faban.client.FabanClient;
-
 import com.mongodb.MongoClient;
-
 import de.thomaskrille.dropwizard_template_config.TemplateConfigBundle;
 import de.thomaskrille.dropwizard_template_config.TemplateConfigBundleConfiguration;
 import io.dropwizard.Application;
@@ -22,11 +20,8 @@ import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import io.federecio.dropwizard.swagger.SwaggerBundle;
 import io.federecio.dropwizard.swagger.SwaggerBundleConfiguration;
-
 import java.util.concurrent.ExecutorService;
-
 import javax.ws.rs.client.Client;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplication.java
+++ b/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplication.java
@@ -6,6 +6,7 @@ import cloud.benchflow.experimentmanager.services.external.BenchFlowTestManagerS
 import cloud.benchflow.experimentmanager.services.external.DriversMakerService;
 import cloud.benchflow.experimentmanager.services.external.FabanManagerService;
 import cloud.benchflow.experimentmanager.services.external.MinioService;
+import cloud.benchflow.experimentmanager.services.external.test.FabanManagerServiceMock;
 import cloud.benchflow.experimentmanager.services.internal.dao.BenchFlowExperimentModelDAO;
 import cloud.benchflow.experimentmanager.services.internal.dao.TrialModelDAO;
 import cloud.benchflow.experimentmanager.scheduler.ExperimentTaskScheduler;
@@ -144,9 +145,15 @@ public class BenchFlowExperimentManagerApplication
     trialModelDAO = new TrialModelDAO(mongoClient);
 
     minioService = configuration.getMinioServiceFactory().build();
-    fabanManagerService = new FabanManagerService(fabanClient);
     driversMakerService = configuration.getDriversMakerServiceFactory().build(client);
     testManagerService = configuration.getTestManagerServiceFactory().build(client);
+
+    // start in test mode or not
+    if (configuration.getTestModeFactory().isMockFaban()) {
+      fabanManagerService = new FabanManagerServiceMock(fabanClient);
+    } else {
+      fabanManagerService = new FabanManagerService(fabanClient);
+    }
 
     experimentTaskScheduler = new ExperimentTaskScheduler(experimentTaskExecutorService);
 
--- a/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
+++ b/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
@@ -1,0 +1,110 @@
+package cloud.benchflow.experimentmanager.services.external.test;
+
+import cloud.benchflow.experimentmanager.constants.BenchFlowConstants;
+import cloud.benchflow.experimentmanager.services.external.FabanManagerService;
+import cloud.benchflow.experimentmanager.tasks.running.execute.ExecuteTrial.TrialStatus;
+import cloud.benchflow.faban.client.FabanClient;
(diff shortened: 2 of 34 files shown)


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplication.java
```diff
@@ -144,10 +140,16 @@ public void run(BenchFlowExperimentManagerConfiguration configuration, Environme
     trialModelDAO = new TrialModelDAO(mongoClient);
 
     minioService = configuration.getMinioServiceFactory().build();
-    fabanManagerService = new FabanManagerService(fabanClient);
     driversMakerService = configuration.getDriversMakerServiceFactory().build(client);
     testManagerService = configuration.getTestManagerServiceFactory().build(client);
 
+    // start in test mode or not
```
VincenzoFerme, line 146: See: https://github.com/benchflow/benchflow/issues/385#issuecomment-306215813

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com) created on 2017-06-04
+ */
+public class FabanManagerServiceMock extends FabanManagerService {
+
+  private static final String SCENARIO_ALWAYS_COMPLETED = "alwayscompleted";
```
VincenzoFerme: We should use the word separator also here. So it should be `always_completed`.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com) created on 2017-06-04
+ */
+public class FabanManagerServiceMock extends FabanManagerService {
```
VincenzoFerme, line 20: Make it explicit all the scenarios you describe in the following comment: https://github.com/benchflow/benchflow/pull/392#issue-233463626

benchflow-experiment-manager/configuration.yml
```diff
@@ -85,4 +85,7 @@ testManager:
 
 experimentTaskExecutor:
   minThreads: 5
-  maxThreads: 15
\ No newline at end of file
+  maxThreads: 15
+
+testMode:
+  mockFaban: ${env.MOCK_FABAN!false?c}
```
VincenzoFerme, line 91: What is `?c`?

benchflow-experiment-manager/application/src/test/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplicationTestModeIT.java
```diff
@@ -0,0 +1,231 @@
...
+        .path(BenchFlowExperimentResource.RUN_ACTION_PATH).request().post(null);
+
+    Assert.assertEquals(Response.Status.NO_CONTENT.getStatusCode(), response.getStatus());
+
+    BenchFlowExperimentModelDAO experimentModelDAO = new BenchFlowExperimentModelDAO(mongoClient);
+
+    Assert.assertNotNull(experimentModelDAO.getExperimentModel(experimentID));
+
+    // wait long enough for tasks to start to be executed
+    executorService.awaitTermination(2, TimeUnit.SECONDS);
```
VincenzoFerme: This is not safe, you need to actually check if they are started, wait if not and check again. It applies to all the cases.

VincenzoFerme (changes requested):

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplication.java
jeppe-style: I updated the issue

benchflow-experiment-manager/configuration.yml
jeppe-style: It is to parse the value as a boolean. I updated the issue https://github.com/benchflow/benchflow/issues/385

benchflow-experiment-manager/application/src/test/java/cloud/benchflow/experimentmanager/BenchFlowExperimentManagerApplicationTestModeIT.java
jeppe-style: I have opened an issue about this since it also applies to the test-manager https://github.com/benchflow/benchflow/issues/395

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
jeppe-style: It will not work since we read the scenario from the `test name` which is part of the `test ID` and when converted to the Faban compatible ID the `.` are replaced with `_`. If we use it also to define the name we cannot convert back to the original ID.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
jeppe-style: The comment is updated

VincenzoFerme: I mean, it is better if you have all the three scenarios clearly defined in the code and then you consider also one of them a default scenario.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
VincenzoFerme: So, I cannot define a test name as "This_is_my_test"? In this case "always_complete"?

jeppe-style: that is correct - it something that was already in the code when I started. In the same way `/` cannot be used since we use that when storing on minio.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
jeppe-style: If I add the text from  https://github.com/benchflow/benchflow/pull/392#issue-233463626 as a comment then it is fine?

VincenzoFerme: No because you have only `SCENARIO_ALWAYS_COMPLETED` and `SCENARIO_FAIL_FIRST_EXECUTION`. I would like to have all the three scenario clearly specified at the top of the class and used in the switch, even though one is also the default one.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
VincenzoFerme: Ok, fine then please use Camel Notation at least: `alwaysComplete`. And please open an issue about the characters that are not allowed in the name of the test (that probably should be checked when accepting the test bundle in the test manager).

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
jeppe-style: ok - I updated https://github.com/benchflow/benchflow/pull/392#issue-233463626 and can add that also to the code

VincenzoFerme: I would make `alwaysCompleted` as default.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
jeppe-style: I wanted to support both camelCase and only lower case and thus convert to received test name to lower case when using in the switch statement. If we define the strings as camelCase I cannot convert them and use it in the switch statement. I can add a comment about that if you agree.

jeppe-style: here is the issue about validating the test name https://github.com/benchflow/benchflow/issues/399

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/test/FabanManagerServiceMock.java
jeppe-style: ok - I changed that in the https://github.com/benchflow/benchflow/pull/392#issue-233463626

VincenzoFerme (approved):

Issue #401, opened 2017-06-10
Update DSL optional (for the Model) values


jeppe-style:
In order to match what is implemented in the thesis ([slack discussion](https://benchflow.slack.com/conversation/G5GCXP63D/p1497120251698733)) change the following:
- [x] workload execution should not be optional
- [x] termination criteria should not be optional (experiment needs to be defined, other optional)
- [x] data collection for client side faban should not be optional
- [x] number of users should not be optional for experiment

VincenzoFerme:
@jeppe-style: which are the ones that are optional also for the user?

jeppe-style:
The ones that have default values as specified here https://github.com/VincenzoFerme/benchflow-temporary-documentation/blob/master/roadmap/Jesper/dsl.md

For this PR:
- data collection for client side faban

PR #439, opened 2017-06-20
Updates optional values

jeppe-style:
Fixes https://github.com/benchflow/benchflow/issues/401

### DIFF
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
@@ -16,9 +16,11 @@ import scala.util.Try
  */
 object BenchFlowExperimentConfigurationYamlProtocol extends DefaultYamlProtocol {
 
-  // default value since deserializing from a test definition it might not be defined
-  // before use it will be overridden
-  val defaultUsers = 0
+  // We use a placeholder since we need to have users defined in the experiment but it may not be defined in
+  // the test that is used to generate the template experiment. The template experiment is then changed according
+  // to the configuration of the exploration point.
+  // During the semantic check afterwards we check that the value has been changed.
+  val UsersPlaceholder: Int = -1
 
   private def keyString(key: YamlString) = s"${BenchFlowTestConfigurationYamlProtocol.Level}.${key.value}"
 
@@ -33,7 +35,7 @@ object BenchFlowExperimentConfigurationYamlProtocol extends DefaultYamlProtocol 
         users <- deserializationHandler(
           yamlObject.getFields(UsersKey).headOption match {
             case Some(usersKey) => usersKey.convertTo[Int]
-            case None => defaultUsers
+            case None => UsersPlaceholder
           },
           keyString(UsersKey))
 
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
@@ -16,6 +16,10 @@ import scala.util.Try
  */
 object BenchFlowExperimentConfigurationYamlProtocol extends DefaultYamlProtocol {
 
+  // default value since deserializing from a test definition it might not be defined
+  // before use it will be overridden
+  val defaultUsers = 0
+
   private def keyString(key: YamlString) = s"${BenchFlowTestConfigurationYamlProtocol.Level}.${key.value}"
 
   implicit object ExperimentConfigurationReadFormat extends YamlFormat[Try[BenchFlowExperimentConfiguration]] {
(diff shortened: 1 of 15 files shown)


benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
```diff
@@ -16,6 +16,10 @@ import scala.util.Try
  */
 object BenchFlowExperimentConfigurationYamlProtocol extends DefaultYamlProtocol {
 
+  // default value since deserializing from a test definition it might not be defined
+  // before use it will be overridden
+  val defaultUsers = 0
```
VincenzoFerme: It is not clear why this value is here, from the comment above.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
```diff
@@ -27,15 +31,18 @@ object BenchFlowExperimentConfigurationYamlProtocol extends DefaultYamlProtocol
       for {
 
         users <- deserializationHandler(
-          yamlObject.getFields(UsersKey).headOption.map(_.convertTo[Int]),
+          yamlObject.getFields(UsersKey).headOption match {
+            case Some(usersKey) => usersKey.convertTo[Int]
+            case None => defaultUsers
```
VincenzoFerme: This sets it to `0`, but that is not correct.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowTestConfiguration.scala
```diff
@@ -12,6 +12,6 @@ import cloud.benchflow.dsl.definition.configuration.workloadexecution.WorkloadEx
 case class BenchFlowTestConfiguration(
   goal: Goal,
   users: Option[Int],
-  workloadExecution: Option[WorkloadExecution],
+  workloadExecution: WorkloadExecution,
   strategy: Option[ExplorationStrategy],
-  terminationCriteria: Option[BenchFlowTestTerminationCriteria])
+  terminationCriteria: BenchFlowTestTerminationCriteria)
```
VincenzoFerme, line 17: How do you check that just some properties of this element needs to be defined?

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
jeppe-style: When creating an experiment from a test definition (the only way we support) we need to provide a default value for the users since it is optional in the test but not optional in the experiment (we always want it to be defined since otherwise it cannot be executed). The default value is more of a temporary placeholder since it has to be overridden when building the experiment using the configuration of the point in the exploration space. If the exploration space does not define  users then the semantic check should have thrown an error earlier.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
jeppe-style: see above comment

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowTestConfiguration.scala
jeppe-style: The `YamlProtocol` checks each key-value pair and will throw an exception if something is missing.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowExperimentConfigurationYamlProtocol.scala
VincenzoFerme: Ok, then you do not need a default. You just need to assign the value you get from the test, and if there is not value as you say it will fail earlier in the lifecycle.

jeppe-style: The experiment is generated from the test definition through parsing. During this process values not part of the experiment are removed. At a later stage the configuration from the exploration space are used to change the generated experiment to adjust it accordingly. Since in this process the test might not have defined the `users` we need to provide a default value, that is then overridden.

VincenzoFerme: Set `-1` and open an issue to explain how these values are handled later on when generating the actual experiment. Also describe how do we need to handle in semantics validation.

jeppe-style: change `defaultUsers` to `usersPlaceholder` and the value to `-1`

comment: "We use a placeholder since we need to have `users` defined in the experiment but it may not be defined in the test that is used to generate the template experiment. The template experiment is then changed according to the configuration of the exploration point. During the semantic check afterwards we check that the value has been changed."

issue for semantic check: https://github.com/benchflow/benchflow/issues/441

jeppe-style: @VincenzoFerme Please check if comment is alright.

VincenzoFerme: @jeppe-style It is, once you add it in the code I can merge the PR.

jeppe-style: @VincenzoFerme now it is added

VincenzoFerme:
Codacy is stuck. Anyway the PR was fine before: https://www.codacy.com/app/BenchFlow/benchflow/pullRequest?prid=722279

VincenzoFerme (approved):

Issue #437, opened 2017-06-20
Defaults for `data_collection.client_side.faban` if missing


jeppe-style:
We should provide defaults for `data_collection.client_side.faban` if the user did not specify them: `max_run_time` and `interval`

```
data_collection:
  client_side:
    faban:
      max_run_time: { String : <max runtime on Faban for a single trial + "h" (hour)>} # default 6h
      interval: { String : <time + "s" (seconds)> } # default 1s
```

We should provide the default as a case class in `FabanConfiguration`. Needs to consider that one of the two values might be specified, so probably multiple case classes are necessary. Alternatively specify the defaults values and use these in the `YamlProtocol`.

The general idea is to add the default configuration during the deserialization (read) in the `YamlProtocol` files.

The `data_collection` part is optional from the user's point of view so we need to check if it was specified in the `YamlProtocol`. If not specified we add it and then continue down the tree. If it was specified we check if `client_side` was specified in that `YamlProtocol` and so on, until we have specified all.

VincenzoFerme:
To me the possibility to clearly identify where the default values are defined in the code is important, because they might change over time.

PR #443, opened 2017-06-21
Adds setting to enable/disable stored knowledge

jeppe-style:
As per title

Fixes https://github.com/benchflow/benchflow/issues/444

### DIFF
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowTestConfiguration.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/BenchFlowTestConfiguration.scala
@@ -1,6 +1,7 @@
 package cloud.benchflow.dsl.definition.configuration
 
 import cloud.benchflow.dsl.definition.configuration.goal.Goal
+import cloud.benchflow.dsl.definition.configuration.settings.SettingsObject.Settings
 import cloud.benchflow.dsl.definition.configuration.strategy.ExplorationStrategy
 import cloud.benchflow.dsl.definition.configuration.terminationcriteria.BenchFlowTestTerminationCriteria
 import cloud.benchflow.dsl.definition.configuration.workloadexecution.WorkloadExecution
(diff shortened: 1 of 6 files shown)


benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/GoalYamlProtocol.scala
```diff
@@ -18,9 +18,12 @@ import scala.util.Try
 object GoalYamlProtocol extends DefaultYamlProtocol {
 
   val TypeKey = YamlString("type")
+  val StoredKnowledgeKey = YamlString("stored_knowledge")
   val ObservationKey = YamlString("observation")
   val ExplorationSpaceKey = YamlString("exploration_space")
 
+  val StoredKnowledgeDefaultValue = true
```
VincenzoFerme: Is this the final way by which we are going to handle defaults? I would have an object holding default values, so that they are clearly identifiable in the class.

VincenzoFerme (changes requested):
To me this has to be in a `configuration` subsection of the `goal` section, because we might need more values and also for make it clear what that is, as it is for all the rest of the DSL. Moreover I would call it `use_stored_knowledge`.

jeppe-style:
@VincenzoFerme I added the discussion into this issue https://github.com/benchflow/benchflow/issues/444

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/GoalYamlProtocol.scala
jeppe-style: I added it there in the same way as in the previous PR

VincenzoFerme: I added the following comment yesterday after our discussion: https://github.com/benchflow/benchflow/issues/437#issuecomment-309818624. Better to have it in an object, also for the other cases.

jeppe-style:
@VincenzoFerme the pull request has been updated and is ready for review

Please also check how the default value is defined is OK

VincenzoFerme:
Defining the default in the DSL entity to which it applies is fine.

VincenzoFerme (approved):

PR #455, opened 2017-06-22
Fixes issue with not properly removing description if not defined

jeppe-style:
fixes https://github.com/benchflow/benchflow/issues/419

depends on PR https://github.com/benchflow/benchflow/pull/443

Issue #424, opened 2017-06-19
Questions on the DSL after Code Review


VincenzoFerme:
Related to https://github.com/benchflow/benchflow/issues/244. See also https://github.com/benchflow/benchflow/issues/425. Some of the following applies also to other services.

- [x] 1. For what do we need the `jitpack.io` repository?
- [x] 2. When I run `mvn:test`, it says that the tests are skipped
- [ ] 3. The following regex is wrong, should be `(m|g)`: https://github.com/benchflow/benchflow/blob/devel/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/deployment/docker/service/ServiceYamlProtocol.scala#L87
- [x] 4. Why the existence of the `model` folder for the SUT type `WfMS` is not part of the generation of the experiment bundle in `generateExperimentBundle`?: https://github.com/benchflow/benchflow/blob/devel/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
- [x] 5. Why is the following file in the root?: https://github.com/benchflow/benchflow/blob/devel/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/ExplorationSpace.scala
- [x] 6. Why the `errorhandling` is only in the `definition` package? Does not it apply also to `docker.compose`?: https://github.com/benchflow/benchflow/tree/devel/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/errorhandling
- [x] 7. Why the `types` is only in the `definition` package? Does not it apply also to `docker.compose`?: https://github.com/benchflow/benchflow/tree/devel/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/types
- [ ] 8. We should rename `data.collection.service: environment -> configuration`
- [ ] 9. The description is optional, so if not specified it has to be set as "" so that on Faban it is not shown.
- [ ] 10. Use the correct [logging level](https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm)

jeppe-style:
1. For what do we need the `jitpack.io` repository? => not sure, think it was added before I started
2. _When I run mvn:test, it says that the tests are skipped_ => This was changed by you https://github.com/benchflow/benchflow/commit/e094f6297dbf7394a9b48cea17c456b217ba0e99
3. 
[...]

Issue #419, opened 2017-06-18
Description is null in the Generated Experiment if Not Specified in Test


VincenzoFerme:

jeppe-style:
@VincenzoFerme I think this is a bug in the `dsl`

PR #456, opened 2017-06-22
Updates POM to get moultingyaml library from maven central

jeppe-style:
as per title

fixes issue in https://github.com/benchflow/benchflow/issues/424#issuecomment-309496973

depends on https://github.com/benchflow/benchflow/pull/455


VincenzoFerme (commented):
The PR has a commit from another PR: 2d36e7f

Maybe the issue is that you force pushed #419

I Reviewed it, when finalised we need to clean it up before merging.

VincenzoFerme (approved):

PR #458, opened 2017-06-22
Fixes typo in DemoConverterTest

jeppe-style:
as per title

fixes https://github.com/benchflow/benchflow/issues/406

depends on https://github.com/benchflow/benchflow/pull/457

PR #459, opened 2017-06-22
Adds experiment termination criteria type as enum

jeppe-style:
Experiment termination criteria Type was only a `String`. With this PR it is converted into an `enum` so we can validate the type.

depends on https://github.com/benchflow/benchflow/pull/458


VincenzoFerme (commented):
The PR has a commit from another PR: 2d36e7f

Maybe the issue is that you force pushed #419

I Reviewed it, when finalised we need to clean it up before merging.

VincenzoFerme (approved):

Issue #397, opened 2017-06-06
Step function in exploration space definition


jeppe-style:
Exploration space definition using `single value` and `step function`. Environment variables only support adding values in a `List[String]`

For step function we want to only support certain operators (for security, and simplicity):
- `+`
- `*`
- `^` (power) 

e.g. `^2`, `+2`

important to handle unit values, e.g. `0*0` or `1^2`

when calculating we should include the values specified in range, e.g. `range: [1,10]`, `step: +2` becomes `[1, 3, 5, 7, 9, 10]`

Update docs also accordingly

VincenzoFerme:
@jeppe-style: related discussion in https://benchflow.slack.com/archives/C5D75QHBN/p1495541101558402

VincenzoFerme:
@jeppe-style can you update the issue to report how to add the support for a configuration options, following what we discussed for users in #465. Moreover we should also have an issue (or this issue) about:

- to provide flexibility we can in the future add exploration strategies to cover all cases (e.g. invert exploration (worst to best))
- in general there should be specific control for each type of exploration space dimension

VincenzoFerme:
We need also to add `\` (division) and `-` (subtraction).

For the `-`:

- Not allowed step value: `0`
- `range.head < range.tail`
- MonotonicPositiveInt: no `--1`

VincenzoFerme:
When implementing the step for other types other than `users` (`cpu`, `ram`, `environment`), the following discussed in https://github.com/benchflow/benchflow/pull/465 has to be fixed:

- https://github.com/benchflow/benchflow/pull/465#discussion_r123870138
- https://github.com/benchflow/benchflow/pull/465#discussion_r123870137
- https://github.com/benchflow/benchflow/pull/465#discussion_r124347521
- Refactor https://github.com/benchflow/benchflow/pull/465#discussion_r124004306, https://github.com/benchflow/benchflow/pull/465#discussion_r124344073, https://github.com/benchflow/benchflow/pull/465#discussion_r124343131, https://github.com/benchflow/benchflow/pull/465#discussion_r124343261, https://github.com/benchflow/benchflow/pull/465#discussion_r124347282
- Add an abstract function so that subtypes know that they have to implement the function https://github.com/benchflow/benchflow/pull/465#discussion_r124343618
- https://github.com/benchflow/benchflow/pull/465#discussion_r124345801
- https://github.com/benchflow/benchflow/pull/465#discussion_r124346317
[...]

PR #465, opened 2017-06-23
Adds range with step for number of users

jeppe-style:
as per title

Fixes (for number of users) https://github.com/benchflow/benchflow/issues/397

Currently supported for number of users but open for extension to other values

@VincenzoFerme please carefully have a look at `RangeWithStep` because I made some assumptions there when to say it is not a valid specification


benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/IntValuesTest.scala
```diff
@@ -50,4 +50,92 @@ class IntValuesTest extends JUnitSuite {
 
   }
```
VincenzoFerme: Cover also `+0`, `*1`, `^0`, `^1`, and in general all the [Absorbing Elements](https://en.wikipedia.org/wiki/Absorbing_element), [Identity Elements](https://en.wikipedia.org/wiki/Identity_element), and [Inverse Elements](https://en.wikipedia.org/wiki/Inverse_element) of the Operators we Have. Also negative numbers for all the operations. 

I saw we just support integer specification, and for now I think it is fine to cover the most common cases.

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/IntValuesTest.scala
```diff
@@ -50,4 +50,92 @@ class IntValuesTest extends JUnitSuite {
...
+
+    val rangeStepYaml: String =
+      """
+        |range: [2,20]
+        |step: "^1"
+      """.stripMargin
+
+    val terminationCriteria = rangeStepYaml.parseYaml.convertTo[Try[IntValues]]
+
+    Assert.assertTrue(terminationCriteria.isFailure)
```
VincenzoFerme: We should also report back the type of problem

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/ExplorationValuesIntYamlProtocol.scala
```diff
@@ -62,10 +72,48 @@ object ExplorationValuesIntYamlProtocol extends DefaultYamlProtocol {
...
+  private def getRangeList(range: List[Int]): Try[List[Int]] = {
+
+    if (range.size != 2) {
+
+      Failure(new BenchFlowDeserializationException("range must be specified as a List of 2 elements", new Throwable))
 
-    (range.head to range(1)).toList
+    } else {
+
+      Success((range.head to range(1)).toList)
```
VincenzoFerme: Is this the default for a range? `+1` as step? If yes, make it clear in a comment.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/step/RangeWithStep.scala
```diff
@@ -0,0 +1,56 @@
...
+object RangeWithStep {
+
+  def stepList[T](
+    range: List[T],
+    step: T,
+    func: (T, T) => T)(implicit num: Numeric[T]): Try[List[T]] = {
+
+    val list = generateList(List(range.head), step, range(1), func)
+
+    if (list.size <= 2) {
```
VincenzoFerme: Validation should happen also before calling this function, so you can also report to the user what is the issue. Here it is fine to check the result you get is fine.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/ExplorationValuesIntYamlProtocol.scala
```diff
@@ -62,10 +72,48 @@ object ExplorationValuesIntYamlProtocol extends DefaultYamlProtocol {
...
+    if (range.size != 2) {
+
+      Failure(new BenchFlowDeserializationException("range must be specified as a List of 2 elements", new Throwable))
+
+    } else {
+
+      val prefix = step.substring(0, 1)
+
+      val triedValue = Try(step.substring(1).toInt)
+
```
VincenzoFerme: You also need to validate the validity of the entire step definition, as discussed in a subsequent comment on the tests. Do not call and wait for a fail of the `RangeWithStep` function if you already know something it is not valid.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/step/RangeWithStep.scala
```diff
@@ -0,0 +1,56 @@
...
+    list: List[T], step: T, maxValue: T, func: (T, T) => T)(implicit num: Numeric[T]): List[T] = {
+
+    val next = func(list.last, step)
+
+    if (num.equiv(next, list.last)) {
+      // if we don't change value we stop
+      // to avoid infinite recursion
+      list
+
+    } else if (list.size == 1 && num.gt(num.abs(num.minus(maxValue, next)), num.abs(num.minus(maxValue, list.head)))) {
```
VincenzoFerme: This case also, it is fine to have it here for security reason, but can be detected before as discussed in a comment on the test section.

VincenzoFerme (changes requested):

VincenzoFerme:
@jeppe-style I also marked https://github.com/benchflow/benchflow/issues/397 as fixed only for number of users, so that we track the need of extending it to other cases.

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/IntValuesTest.scala
jeppe-style: we do report type of failure but in the test I think it is sufficient for now to just check that it actually failed as expected.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/ExplorationValuesIntYamlProtocol.scala
jeppe-style: added

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/step/RangeWithStep.scala
jeppe-style: I check that the range has two elements in the list, that the step function is valid and that the step itself is an integer. Not sure what we can test beyond that before running. If you have any ideas that would be helpful.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/step/RangeWithStep.scala
jeppe-style: I don't see how that can be done.

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/IntValuesTest.scala
jeppe-style: I added now more tests for the cases you specified. I also added subtraction so that we also can handle decreasing sequences.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/ExplorationValuesIntYamlProtocol.scala
jeppe-style: related to next comment

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/IntValuesTest.scala
```diff
@@ -90,6 +110,26 @@ class IntValuesTest extends JUnitSuite {
...
+
+    val rangeStepYaml: String =
+      """
+        |range: [1,10]
+        |step: "*-2"
+      """.stripMargin
+
+    val terminationCriteria = rangeStepYaml.parseYaml.convertTo[Try[IntValues]]
+
+    Assert.assertTrue(terminationCriteria.isSuccess)
```
VincenzoFerme: Why is this valid? The range should be `a < b` for this to be valid.

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/step/RangeWithStep.scala
VincenzoFerme: Yes, everything that math says is not valid (see https://github.com/benchflow/benchflow/pull/465#discussion_r123855300)

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/step/RangeWithStep.scala
VincenzoFerme: See https://github.com/benchflow/benchflow/pull/465#discussion_r123855300

jeppe-style:
Discussion:
- [x] add the error messages for the errors detected during generation of the list (e.g. return `Try[List[Numeric]]`)
- [x] errors are defined by messages (e.g. no hierarchy necessary for now)
- [x] requirement of the range is that it goes from best expected performance to worse expected performance
- [x] number of users should always be greater than 0
- [x] add validation for math (e.g. identity etc.) before calling `RangeWithStep.stepList`


-  to provide flexibility we can in the future add exploration strategies to cover all cases (e.g. invert exploration (worst to best))
- in general there should be specific control for each type of exploration space dimension

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -52,46 +53,66 @@ object ExplorationValuesIntYamlProtocol extends DefaultYamlProtocol {
...
+      Failure(new BenchFlowDeserializationException("values must be > 0", new Throwable))
+    } else {
+      Success(values)
+    }
 
   }
 
   private def getRangeList(range: List[Int]): Try[List[Int]] = {
 
     if (range.size != 2) {
```
VincenzoFerme: The validation part looks the same as the `getStepList` function: https://github.com/benchflow/benchflow/pull/465/files/1c9cced7a66798daefdb836ab19fc561df17bb7f..561cf5d61e4b9bddd849242517c01b03d56da5c7#diff-5fbdece869476b79c16c33a628a1057dR108

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -111,10 +137,26 @@ object ExplorationValuesIntYamlProtocol extends DefaultYamlProtocol {
...
+
+    case "*" =>
+      if (step == 1) {
+        Failure(new BenchFlowDeserializationException("step function cannot be identity element: *1", new Throwable))
+      } else {
+        RangeWithStep.stepList(range, step, (a: Int, b: Int) => a * b)
+      }
+
+    case "^" =>
+      if (step == 1) {
```
VincenzoFerme: Also `^0` is not valid, because you keep getting `1`.

benchflow-dsl/src/test/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/explorationvalues/IntValuesTest.scala
VincenzoFerme: We are safe from https://en.wikipedia.org/wiki/Inverse_element#Real_numbers for now because we just support integer numbers as steps.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -111,10 +137,26 @@ object ExplorationValuesIntYamlProtocol extends DefaultYamlProtocol {
...
-    case _ => Failure(new BenchFlowDeserializationException("step function not supported", new Throwable))
+    case "+" =>
+      if (step == 0) {
+        Failure(new BenchFlowDeserializationException("step function cannot be identity element: +0", new Throwable))
+      } else {
+        RangeWithStep.stepList(range, step, (a: Int, b: Int) => a + b)
+      }
+
+    case "*" =>
+      if (step == 1) {
```
VincenzoFerme: Also `*0` is not valid because you are keep getting `0` (https://en.wikipedia.org/wiki/Absorbing_element). Place also a reference to the way you perform these checks.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -111,10 +137,26 @@ object ExplorationValuesIntYamlProtocol extends DefaultYamlProtocol {
   }
 
   private def generateStepList(range: List[Int], step: Int, prefix: String): Try[List[Int]] = prefix match {
```
VincenzoFerme: The checks, other than the comments, are fine. But we should check also the seed number (so in our case the first element in the range) for the same values, because that is used as a first argument in the operation.

VincenzoFerme (changes requested):
These are the changes that we still to apply before being able to merge the pull request. Including: https://github.com/benchflow/benchflow/pull/465#pullrequestreview-46255861

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+      Failure(new BenchFlowDeserializationException("values must be increasing", new Throwable))
+    } else if (values.head <= 0) {
+      Failure(new BenchFlowDeserializationException("values must be > 0", new Throwable))
+    } else {
+      Success(values)
+    }
+
+  }
+
+  private def getRangeList(range: List[Int]): Try[List[Int]] = {
```
VincenzoFerme: General checks goes in the superclass, type specific steps have to be here.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+    } else {
+      // IF only range then the step is 1
+      Success((range.head to range(1)).toList)
+    }
+
+  }
+
+  private def getStepList(range: List[Int], step: String): Try[List[Int]] = {
+
+    if (range.size != 2) {
```
VincenzoFerme: General checks goes in the superclass, type specific steps have to be here.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+
+      Map[YamlValue, YamlValue](
+        ValuesKey -> obj.values.toYaml)
+
+    }
+
+    override def read(yaml: YamlValue): UserValues = unsupportedReadOperation
+
+  }
+
```
VincenzoFerme, line 71: The following function should be enforced from the superclass, so that when implementing a new type, it is clear they have to be defined. E.g., define them `abstract`.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+  private def assertValidValues(values: List[Int]): Try[List[Int]] = {
+
+    // check that all values are in increasing order
```
VincenzoFerme: General checks goes in the superclass, type specific steps have to be here.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+    val increasingOrder = values.sliding(2).forall {
+      case List(x, y) => x < y
+    }
+
+    if (!increasingOrder) {
+      Failure(new BenchFlowDeserializationException("values must be increasing", new Throwable))
```
VincenzoFerme: When you report this `Failure` is it clear in which part of the DSL definition there was the issue?

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+  private def getRangeList(range: List[Int]): Try[List[Int]] = {
+
+    if (range.size != 2) {
+      Failure(new BenchFlowDeserializationException("range must be specified as a List of 2 elements", new Throwable))
+    } else if (range.head >= range.last) {
+      Failure(new BenchFlowDeserializationException("range must be increasing", new Throwable))
+    } else if (range.head <= 0) {
+      Failure(new BenchFlowDeserializationException("range must be > 0", new Throwable))
+    } else {
+      // IF only range then the step is 1
```
VincenzoFerme: This is a default we implicitly assume, make it clear in the class definition as we did for the other default values.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
VincenzoFerme: Moreover most of the checks are duplicated with https://github.com/benchflow/benchflow/pull/465/files#diff-5fbdece869476b79c16c33a628a1057dR95

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+      Map[YamlValue, YamlValue](
+        ValuesKey -> obj.values.toYaml)
+
+    }
+
+    override def read(yaml: YamlValue): UserValues = unsupportedReadOperation
+
+  }
+
+  private def assertValidValues(values: List[Int]): Try[List[Int]] = {
```
VincenzoFerme: You can define a `MonotonicPositiveInt` type to check for this, extending a `MonotonicInt` type that just check for https://github.com/benchflow/benchflow/pull/465/files#diff-5fbdece869476b79c16c33a628a1057dR79. Then `MonotonicInt` can extend `Int` (see https://github.com/benchflow/benchflow/pull/465/files#diff-5fbdece869476b79c16c33a628a1057dR118). This way you increase reusability, especially for these cases that are very common in our DSL.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    if (range.size != 2) {
+      Failure(new BenchFlowDeserializationException("range must be specified as a List of 2 elements", new Throwable))
+    } else if (range.head >= range.last) {
+      Failure(new BenchFlowDeserializationException("range must be increasing", new Throwable))
+    } else if (range.head <= 0) {
+      Failure(new BenchFlowDeserializationException("range must be > 0", new Throwable))
+    } else {
+
+      val prefix = step.substring(0, 1)
```
VincenzoFerme: Also for this you can have a type, say `PositiveIntStep` extending `IntStep` extending a general type `Int`.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+        case Failure(ex) => Failure(new BenchFlowDeserializationException("step must contain an integer", ex))
+
+      }
+
+    }
+
+  }
+
+  private def generateStepList(range: List[Int], step: Int, prefix: String): Try[List[Int]] = prefix match {
+    case "+" =>
```
VincenzoFerme: All the checks are generic and can go in the superclass, only the actual function is specific but it is always the same given the type and can be called by the supertype.

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
```diff
@@ -0,0 +1,162 @@
...
+import cloud.benchflow.dsl.definition.errorhandling.YamlErrorHandler.{ deserializationHandler, unsupportedReadOperation, unsupportedWriteOperation }
+import net.jcazevedo.moultingyaml.{ DefaultYamlProtocol, YamlFormat, YamlObject, YamlString, YamlValue, _ }
+
+import scala.util.{ Failure, Success, Try }
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com)
+ *         created on 2017-05-23
+ */
+object UserValuesYamlProtocol extends DefaultYamlProtocol {
```
VincenzoFerme: You deleted https://github.com/benchflow/benchflow/pull/465/files#diff-5b7b9c50538d06a01fd7c3dca9fa2dd4 and added this class. This class has much code that is similar to other classes (or it is going to be similar).

According to me we should:

- Have base types (equivalent to primitive types in a programming language), that are defined independently of the DSL entity so that can be reused to obtain values. In this case `Int` would be an independent type, and `User` would use it. This way you can use the same definition to get the value of `users` independently of where in the DSL it is defined (since we allow it to be defined as exploration parameter or as configuration. 
- Have a generic definition for shared elements, as `values`, `range`, `step` that can be reused for all the elements needing it. This generic definition is reused, and according to the context (e.g., reading the `users` or the `memory` definition) the correct Reader can be called. 


Update this PR or open an issue to describe the needed changes.

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: It is similar but not exactly the same. I wanted it to be explicit therefore I did not put it into a common method.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: I check this already in the in the `getStepList` method. I remove the check from the `+` case to  make it consistent.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: I added the check for the `^` case. For other cases there is no impact.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: I added a comment that it is checked prior to calling this function

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: I did not do that because for each type the allowed ranges might be different. I think it is better to do this refactoring once there is a need. I need now to focus on finishing for the demo.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: It was not. I have fixed this now.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: It is not specific to this class but in general therefore I wouldn't put it in the class definition.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/configuration/goal/explorationspace/workload/users/UserValuesYamlProtocol.scala
jeppe-style: see https://github.com/benchflow/benchflow/pull/465#discussion_r124842226

VincenzoFerme:
I moved in https://github.com/benchflow/benchflow/issues/397#issuecomment-314020445, what can be implemented when extending the support to other types, what is left for this pull request it:

- https://github.com/benchflow/benchflow/pull/465#discussion_r123870137
- The same kind of checks should happen all in the same place https://github.com/benchflow/benchflow/pull/465#discussion_r124005724, https://github.com/benchflow/benchflow/pull/465#discussion_r124006556
- Check that this is done for all the operators https://github.com/benchflow/benchflow/pull/465#discussion_r124008640

VincenzoFerme:
@jeppe-style I implemented https://github.com/benchflow/benchflow/pull/465#issuecomment-314020453. Can you please review the pull request for the changes I have made.

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+
+      val (prefix: String, stepValue: Int) = getIntStep(step)
+
+      val tryStepList = generateStepList(range, stepValue, prefix)
+
+      if(tryStepList.isSuccess)
+        Success(tryStepList.get)
+      else
+        Failure(tryStepList.failed.get)
+
```
jeppe-style: I think in these lines we can directly return `generateStepList` as it has the same return type

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    }
+
+  }
+
+  protected def getIntStep(step: String) = {
+
+    val (prefix: String, triedStepValue: Try[Int]) = tryGetStep(step)
+
```
jeppe-style: It seems this method is only called once so we could integrate it into this method, also because we explicitly get an int step in the `tryGetStep` method

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+   */
+  protected def generateStepList(range: List[Int], stepValue: Int, prefix: String): Try[List[Int]] = prefix match {
+
+    case "+" =>
+      val increasingRange = assertIncreasingRange(range)
+
+      if(stepValue == 0) {
+        Failure(BenchFlowDeserializationExceptionMessage("step function cannot be +0"))
+      } else if (increasingRange.isFailure) {
+        Failure(increasingRange.failed.get)
```
jeppe-style: we can directly return the `increasingRange` as it is already a `Failure`

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+   * We mostly check for Absorbing, Identity, Inverse and Negative elements
+   *
+   * NOTE: we might have some redundancy with checks in other methods, but this enhance the
+   * reusability of this method and make it generic for all int cases
+   */
+  protected def generateStepList(range: List[Int], stepValue: Int, prefix: String): Try[List[Int]] = prefix match {
+
+    case "+" =>
+      val increasingRange = assertIncreasingRange(range)
+
```
jeppe-style, line 55: The value inside the `Try` is never used

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+    case "*" =>
+      val increasingRange = assertIncreasingRange(range)
+
+      if(stepValue == 0 || stepValue == 1){
+        Failure(BenchFlowDeserializationExceptionMessage("step function cannot be *0 or *1"))
+      } else if (range.head == 0) {
+        Failure(BenchFlowDeserializationExceptionMessage("first element in the list cannot be *0"))
+      } else if (increasingRange.isFailure) {
+        Failure(increasingRange.failed.get)
```
jeppe-style: we can directly return the `increasingRange` as it is already a `Failure`

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+
+      if(stepValue == 0) {
+        Failure(BenchFlowDeserializationExceptionMessage("step function cannot be +0"))
+      } else if (increasingRange.isFailure) {
+        Failure(increasingRange.failed.get)
+      } else
+        RangeWithStep.stepList(range, stepValue, (a: Int, b: Int) => a + b)
+
+    case "*" =>
+      val increasingRange = assertIncreasingRange(range)
```
jeppe-style: The value inside the `Try` is never used

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+      } else if (range.head == 0) {
+        Failure(BenchFlowDeserializationExceptionMessage("first element in the list cannot be *0"))
+      } else if (increasingRange.isFailure) {
+        Failure(increasingRange.failed.get)
+      } else {
+        RangeWithStep.stepList(range, stepValue, (a: Int, b: Int) => a * b)
+      }
+
+    case "^" =>
+      val increasingRange = assertIncreasingRange(range)
```
jeppe-style: The value inside the `Try` is never used

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/IntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,103 @@
...
+
+      if(stepValue <= 0 || stepValue == 1){
+        Failure(BenchFlowDeserializationExceptionMessage("step function cannot be ^-N, ^0 or ^1"))
+      } else if (range.head == 0 || range.head == 1) {
+        Failure(BenchFlowDeserializationExceptionMessage("first element in the list cannot be ^0 or ^1"))
+      } else if (increasingRange.isFailure) {
+        Failure(increasingRange.failed.get)
```
jeppe-style: we can directly return the `increasingRange` as it is already a `Failure`

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/MonotonicPositiveIntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  /*
+   * We must ensure that we can not get negative values when applying the step
+   */
+  protected override def getStepList(range: List[Int], step: String): Try[List[Int]] = {
+
+    val rangeValid = assertRangeIsValid(range)
+
+    if (rangeValid.isFailure) {
+      Failure(rangeValid.failed.get)
```
jeppe-style: we can directly return the `rangeValid` as it is already a `Failure`

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/definition/yamlprotocol/valuesrange/MonotonicPositiveIntValuesRangeYamlProtocol.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    if (rangeValid.isSuccess) {
+      //Positive
+      if (range.head <= 0) {
+        Failure(BenchFlowDeserializationExceptionMessage("range head value must be > 0"))
+      } else {
+        Success(range)
+      }
+    } else {
+      Failure(rangeValid.failed.get)
```
jeppe-style: we can directly return the `rangeValid` as it is already a Failure

jeppe-style (commented):
I had a quick look now and added some comments. To me it seems there is some repetition also in some places since there is some nesting in the traits.

VincenzoFerme:
@jeppe-style I committed an updated version where I apply your comments. We can deal with repetitions when implementing https://github.com/benchflow/benchflow/issues/397#issuecomment-314408192

jeppe-style:
@VincenzoFerme ok for me. I cannot approve since I'm the original author.

VincenzoFerme (approved):

Issue #445, opened 2017-06-21
Add links in REST API response objects


jeppe-style:
**Run BenchFlow Test**
- [x] link to test status (`status: URL`)

**Test Status**
Test Model
- [x] link to test bundle on minio (`testBundle: URL`)

Experiment Model
- [x] link to experiment bundle on minio (`experimentBundle: URL`)
- [x] link to collected data on minio (`collectedData: URL`)
- [x] link to experiment status on experiment-manager (`status: URL`)

Related to https://github.com/benchflow/benchflow/issues/128

VincenzoFerme:
Since these link can change, we need to carefully design a place in the system where the information they are based on, can be specified. The Base URL has to be anyway retrieved from the configuration file I think.

The link to Minio should be to the container "folder" because otherwise I think they just download the file instead of visualising it (that can also be ok).

On Minio we do not have results, but collected data. There will be a link to results, that is going to be probably a link to a Notebook preloading a template to access and query the data, for now.

jeppe-style:
ok - I updated and wrote `link to collected data on minio`

for the rest it is fine?

VincenzoFerme:
The drop wizard configuration can express the reference to the services using an hostname, that is resolved using the DNS service of the infrastructure on which we deploy BenchFlow, in our case currently Rancher. That said, you should determine if the specified server is a valid and reachable IP address, or you need to [convert it to a valid IP](https://stackoverflow.com/questions/15788453/resolving-ip-address-of-a-hostname). This should happen when you load the configuration information, by adding extra fields to the configurations made available to the service.

VincenzoFerme:
TODOs:
- Moreover, do not save URLs to the database, they need to be generated on the fly and they should be positioned in a dedicated section in the response, not mixed with other fields, see: https://github.com/benchflow/benchflow/issues/128#issuecomment-313679357.. The fields should only contain human readable data, so if you want to extend some of them with actual content feel free to do so.
- `collectedData: URL` misses full path (needs hash, not sure of what)

VincenzoFerme:
What is left from this Issue is now documented in https://github.com/benchflow/benchflow/issues/128#issuecomment-319903870

Issue #128, opened 2017-04-10
Add HATEOS for RESTFUL interactions


jeppe-style:
Add [HATEOS](https://en.wikipedia.org/wiki/HATEOAS) where needed (specially for client) to easily navigate resources (e.g. get a list of all experiments with a given name)

- provide links to minio where data can be found (e.g. test bundles, raw performance data)
- provide links to cassandra (or queries) how to retrieve computed metrics

VincenzoFerme:
When applying this, we should also evaluate https://github.com/benchflow/benchflow/issues/102 and callbacks based interaction.

VincenzoFerme:
See link headers: https://cpoepke.wordpress.com/2014/03/06/true-restful-hateoas-for-dropwizard/

VincenzoFerme:
`collectedData`: URL misses full path (needs hash, not sure of what)

VincenzoFerme:
See the following comments, containing improvements that need to be applied while working on HATEOS:

- https://github.com/benchflow/benchflow/issues/445#issuecomment-310144934
- https://github.com/benchflow/benchflow/issues/445#issuecomment-314056238
- https://github.com/benchflow/benchflow/issues/445#issuecomment-316657447

PR #469, opened 2017-06-30
Adds links response objects

jeppe-style:
as per title

fixes partially https://github.com/benchflow/benchflow/issues/445

- `collectedData: URL` misses full path (needs hash, not sure of what)

What is left in this PR, is documented in https://github.com/benchflow/benchflow/issues/445#issuecomment-316657447 for future improvement.


VincenzoFerme (changes requested):
See https://github.com/benchflow/benchflow/issues/445#issuecomment-314056238.

Moreover, do not save URLs to the database, they need to be generated on the fly and they should be positioned in a dedicated section in the response, not mixed with other fields, see: https://github.com/benchflow/benchflow/issues/128#issuecomment-313679357. The fields should only contain human readable data, so if you want to extend some of them with actual content feel free to do so.

jeppe-style:
@VincenzoFerme I think your comments can be added to an issue. Right now it works and serves the purpose of the demo and use cases.

VincenzoFerme (approved):

VincenzoFerme:
What is left in this PR, is documented in https://github.com/benchflow/benchflow/issues/445#issuecomment-316657447 for future improvement.

PR #470, opened 2017-07-01
Adds links to response objects

jeppe-style:
as per title

fixes https://github.com/benchflow/benchflow/issues/446

PR #471, opened 2017-07-01
Updates DSL version

jeppe-style:
as per title

depends on PR https://github.com/benchflow/benchflow/pull/470

@VincenzoFerme locally it builds with wercker - don't understand why it fails here. In the next PR (https://github.com/benchflow/benchflow/pull/472) it works again. Fixed in PR https://github.com/benchflow/benchflow/pull/473.


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/start/StartTask.java
```diff
@@ -60,9 +60,10 @@ public Boolean call() throws Exception {
           .toString(minioService.getExperimentDefinition(experimentID), StandardCharsets.UTF_8);
 
       BenchFlowExperiment experiment =
-          BenchFlowDSL.experimentFromExperimentYaml(experimentYamlString);
+          BenchFlowExperimentAPI.experimentFromExperimentYaml(experimentYamlString);
 
-      int numTrials = experiment.configuration().terminationCriteria().get().experiment().number();
+      int numTrials =
```
VincenzoFerme, line 65: What happens if this is not defined?

VincenzoFerme (changes requested):

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/tasks/start/StartTask.java
jeppe-style: It has to be defined otherwise the DSL library will throw an exception

VincenzoFerme (approved):

PR #472, opened 2017-07-01
Updates handling of Faban results

jeppe-style:
as per title

fixes https://github.com/benchflow/benchflow/issues/440

depends on PR https://github.com/benchflow/benchflow/pull/471

Issue #409, opened 2017-06-14
Illegal run status (unknown) from Faban


jeppe-style:
during execution of an experiment following error occurred and stopped the progress

```
benchflow-experiment-manager_1  | java.util.concurrent.ExecutionException: cloud.benchflow.faban.client.exceptions.IllegalRunStatusException: RunId jespertest_TestExploreUsersOnCamunda750_5_3.1Ureturned illegal run status UNKNOWN
benchflow-experiment-manager_1  | 
benchflow-experiment-manager_1  | 	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
benchflow-experiment-manager_1  | 	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
benchflow-experiment-manager_1  | 	at cloud.benchflow.experimentmanager.scheduler.ExperimentTaskScheduler.handleDetermineAndExecuteTrials(ExperimentTaskScheduler.java:193)
benchflow-experiment-manager_1  | 	at cloud.benchflow.experimentmanager.scheduler.ExperimentTaskScheduler.handleRunningState(ExperimentTaskScheduler.java:156)
benchflow-experiment-manager_1  | 	at cloud.benchflow.experimentmanager.scheduler.ExperimentTaskScheduler.handleExperimentState(ExperimentTaskScheduler.java:95)
benchflow-experiment-manager_1  | 	at cloud.benchflow.experimentmanager.scheduler.ExperimentDispatcher.run(ExperimentDispatcher.java:69)
benchflow-experiment-manager_1  | 	at java.lang.Thread.run(Thread.java:748)
benchflow-experiment-manager_1  | Caused by: cloud.benchflow.faban.client.exceptions.IllegalRunStatusException: RunId jespertest_TestExploreUsersOnCamunda750_5_3.1Ureturned illegal run status UNKNOWN
```

We should handle the `UNKNOWN` case without any runtime exception in [RunStatus.java](https://github.com/benchflow/benchflow/blob/devel/benchflow-faban-client/src/main/java/cloud/benchflow/faban/client/responses/RunStatus.java)

VincenzoFerme:
Fix considering moving to managed exceptions:

- https://10kloc.wordpress.com/2013/03/09/runtimeexceptions-try-catch-or-not-to-catch/
- http://www.javapractices.com/topic/TopicAction.do?Id=129

VincenzoFerme:
[...]

PR #473, opened 2017-07-02
Handles illegal run status (UNKNOWN) and Faban Client exceptions

jeppe-style:
as per title

related to https://github.com/benchflow/benchflow/issues/409

depends on PR https://github.com/benchflow/benchflow/pull/472

### DIFF
--- a/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
+++ b/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
@@ -211,6 +211,7 @@ public class FabanManagerService {
       e.printStackTrace();
     }
 
+    //See https://github.com/benchflow/benchflow/pull/473/files#r128371872
     return new FabanStatus(trialID, UNKNOWN, Result.UNKNOWN);
 
   }
--- a/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
+++ b/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
@@ -4,17 +4,28 @@ import static cloud.benchflow.experimentmanager.constants.BenchFlowConstants.MOD
 import static cloud.benchflow.faban.client.responses.RunStatus.StatusCode.QUEUED;
 import static cloud.benchflow.faban.client.responses.RunStatus.StatusCode.RECEIVED;
 import static cloud.benchflow.faban.client.responses.RunStatus.StatusCode.STARTED;
+import static cloud.benchflow.faban.client.responses.RunStatus.StatusCode.UNKNOWN;
 
 import cloud.benchflow.experimentmanager.BenchFlowExperimentManagerApplication;
 import cloud.benchflow.experimentmanager.constants.BenchFlowConstants;
 import cloud.benchflow.experimentmanager.tasks.running.execute.ExecuteTrial.FabanStatus;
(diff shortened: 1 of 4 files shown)


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
```diff
@@ -153,32 +172,46 @@ public FabanStatus pollForTrialStatus(String trialID, RunId runId) throws RunIdN
...
+          || status.getStatus().equals(STARTED) || status.getStatus().equals(UNKNOWN)) {
 
-    RunInfo runInfo = fabanClient.runInfo(runId);
+        try {
+          Thread.sleep(30 * 1000);
+        } catch (InterruptedException e) {
+          e.printStackTrace();
+        }
+
+        status = fabanClient.status(runId);
```
VincenzoFerme, line 197: You can end up here also if you have an exception, and that should not be the case.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
```diff
@@ -153,32 +172,46 @@ public FabanStatus pollForTrialStatus(String trialID, RunId runId) throws RunIdN
...
+        | FabanClientBadRequestException e) {
+      // TODO - handle me
+      e.printStackTrace();
+    } catch (IllegalRunInfoResultException e) {
+      // TODO - handle me
+      e.printStackTrace();
+    }
 
-    return new FabanStatus(trialID, status.getStatus(), runInfo.getResult());
+    return new FabanStatus(trialID, UNKNOWN, Result.UNKNOWN);
```
VincenzoFerme: Why do you have this, instead of the actual value you might have got before? Moreover `UNKNOWN` is a valid state Faban could return, here you are using it more like a default internal state of BenchFlow when you do not know what to set.

VincenzoFerme (changes requested):

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
jeppe-style: not sure what you mean - this code was accepted earlier

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
jeppe-style: I do return earlier the actual value. This is in the case there was some exception since they are not currently handled.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
VincenzoFerme: It is fine, you query again for the status and just log the exception

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/FabanManagerService.java
VincenzoFerme: All right, makes sense. We simply assume that if we get an exception, we do not know the state. Then in the lifecycle we handle this case with a trial re-execution due to a `RANDOM_FAILURE`.

VincenzoFerme (approved):

Issue #281, opened 2017-05-04
Termination Criteria: To Be Implemented and Impact on Selection Strategies


VincenzoFerme:
Test Termination Criteria:

- `max_time`: maximum execution time for the test

Impact of Selection Strategies: 

- `heuristics`: selection of the part of the space to explore given the time constraints, according to the defined objective @jeppe-style said that this can e itself an exploration strategy that then relies and select other exploration strategies dynamically.

Experiment Termination Criteria:

- `fixed`: can be decided beforehand
- `confidence_interval`: take into account in the design and implementation, because it requires to be verified every trial

jeppe-style:
First implementation with tests.

jeppe-style:
test termination criteria implemented on TM in PR https://github.com/benchflow/benchflow/pull/474

PR #474, opened 2017-07-02
Adds max running time termination criteria

### LINKED ISSUE #337: Test Termination Criteria - Max Time
```
configuration:
  termination_criteria:
    test:
      max_running_time: { String : <amount of time + unit> }
```
When specified in the test definition there should be a timeout triggered after the specified time has passed so that the test can be stopped (any running experiments must also be stopped) and set to state `Terminated` `Partially Complete`.

@VincenzoFerme when do we start to count this time? When the test has ben submitted by the user and become part the lifecycle? Another option is to only count the time it is actually running (state=running), e.g., we do not count the waiting time. => Count only the running time

We can use a [ScheduledExecutorService](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html) to schedule a `Runnable` (`TestTimeout`) that notifies the Scheduler that the termination criteria has been reached after a given delay. 

When submitting the `Runnable` we get a `Future` that can be stored in a `ConcurrentMap<String Future>` (key=testID). In this way we can remove the `TestTimeout` if the test completes before the max time.

The `ScheduledExecutorService` can be instantiated in the application and fetched in the same way as the `TaskExecutorService`.

jeppe-style:
as per title
- Adds method to terminate experiment in task scheduler
- Refactors running states handling to separate class
- Adds morphia converter for Time

fixes https://github.com/benchflow/benchflow/issues/337 (except extensions, that are now in https://github.com/benchflow/benchflow/issues/517)

related to https://github.com/benchflow/benchflow/issues/281

depends on PR https://github.com/benchflow/benchflow/pull/469

TODO

- [x] issues with calling wrong function: [TestTaskScheduler.java#L273](https://github.com/benchflow/benchflow/blob/dee2a5c78260696452975cdd19bb5c62c51c089f/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java#L273), [TestTaskScheduler.java#L281 ](https://github.com/benchflow/benchflow/blob/dee2a5c78260696452975cdd19bb5c62c51c089f/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java#L281)
- [x] issues that not all states (except running) check if test is terminated.


jeppe-style:
issues with calling wrong function: https://github.com/benchflow/benchflow/blob/dee2a5c78260696452975cdd19bb5c62c51c089f/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java#L273

https://github.com/benchflow/benchflow/blob/dee2a5c78260696452975cdd19bb5c62c51c089f/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java#L281

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/BenchFlowTestManagerApplication.java
```diff
@@ -127,6 +128,8 @@ public void run(BenchFlowTestManagerConfiguration configuration, Environment env
     // http://mongodb.github.io/mongo-java-driver/3.4/driver/getting-started/quick-start/
     MongoClient mongoClient = configuration.getMongoDBFactory().build();
     ExecutorService taskExecutor = configuration.getTaskExecutorFactory().build(environment);
+    ScheduledThreadPoolExecutor timeOutScheduledThreadPoolExecutor =
```
VincenzoFerme: You should build it as you do for the other executor service we have: https://github.com/benchflow/benchflow/pull/474/files#diff-3ebdba5e745234429a46f7536fa3feb8R130

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -172,44 +189,51 @@ private synchronized void handleTestRunningState(String testID) {
...
+          runningStatesHandler.handleExperimentResult(testID);
           break;
 
         case VALIDATE_TERMINATION_CRITERIA:
-          validateTerminationCriteria(testID);
+          runningStatesHandler.validateTerminationCriteria(testID);
           break;
 
         case DERIVE_PREDICTION_FUNCTION:
           testModelDAO.setTestRunningState(testID, VALIDATE_PREDICTION_FUNCTION);
```
VincenzoFerme: Why you do this here and not in the function called in the next line?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -135,6 +135,11 @@ private synchronized void handleStartState(String testID) {
       // wait for task to complete
       future.get();
 
+      if (isTerminated(testID)) {
```
VincenzoFerme: Why don't you check this in the `setTestState` method, so that is done only in one place? 

Anyway this should not be possible, because once you get a timeout, you need to get exclusive access to the data structure where you keep running information and cleanup everything before other code gets executed.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -156,6 +161,18 @@ private synchronized void handleWaitingState(String testID) {
       // set state as ready
       testModelDAO.setTestState(testID, BenchFlowTestState.READY);
 
+      // update max running time timeout
```
VincenzoFerme: Why do you need the following code? Is it to restart the execution, and take into account of the time the test has already been executed? If yes, I think this should happen in the `RUNNING` state when the test goes back to `RUNNING`.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -172,44 +189,51 @@ private synchronized void handleTestRunningState(String testID) {
       logger
           .info("handleTestRunningState for " + testID + " with state " + testRunningState.name());
 
+      // set timeout if not already set
+      if (!timeoutTasks.containsKey(testID)) {
```
VincenzoFerme: What happens if the test has no timeout?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -172,44 +189,51 @@ private synchronized void handleTestRunningState(String testID) {
...
+        Time maxRunTime = testModelDAO.getMaxRunningTime(testID);
+        TimeoutTask timeoutTask = new TimeoutTask(testID, this);
+        ScheduledFuture<?> timeoutFuture = timeoutScheduledThreadPoolExecutor.schedule(timeoutTask,
+            maxRunTime.toSecondsPart(), TimeUnit.SECONDS);
+        timeoutTasks.put(testID, timeoutFuture);
+      }
```
VincenzoFerme: How do you make sure you start the timeout in parallel to the execution of the test?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
-
-  private void removeNonReachableExperiments(String testID) {
-
-    logger.info("removeNonReachableExperiments with testID: " + testID);
-
-    RemoveNonReachableExperimentsTask task = new RemoveNonReachableExperimentsTask(testID);
+  private void handleTerminatedState(String testID) {
+    // remove max running time timeout
+    ScheduledFuture timeoutTaskFuture = timeoutTasks.remove(testID);
+    timeoutTaskFuture.cancel(true);
```
VincenzoFerme: This [fail if the task has already be completed](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#cancel(boolean)) so you need to handle the case.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
 
-        case GOAL_NOT_REACHABLE:
-          testModelDAO.setTestState(testID, TERMINATED);
-          testModelDAO.setTestTerminatedState(testID, COMPLETED_WITH_FAILURE);
+      switch (testState) {
+        case READY:
+          // remove from ready queue
+          readyQueue.remove(testID);
+          // set to terminated
+          testModelDAO.setTestState(testID, BenchFlowTestState.TERMINATED);
```
VincenzoFerme: The following two method calls can be refactored in a method, because they are always the same for all the cases.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
+          testModelDAO.setTestTerminatedState(testID, TestTerminatedState.PARTIALLY_COMPLETE);
```
VincenzoFerme: In this case I would not use `PARTIALLY_COMPLETE` because it was not even started.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
-      switch (result) {
+      BenchFlowTestState testState = testModelDAO.getTestState(testID);
 
-        case GOAL_NOT_REACHABLE:
-          testModelDAO.setTestState(testID, TERMINATED);
-          testModelDAO.setTestTerminatedState(testID, COMPLETED_WITH_FAILURE);
+      switch (testState) {
+        case READY:
+          // remove from ready queue
+          readyQueue.remove(testID);
```
VincenzoFerme: How do you handle concurrent access to this data structure? It is implemented by the data structure, but how do you ensure that the execution of the task related to the `READY` state does not interfere with the operation you have here?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
+          testModelDAO.setTestTerminatedState(testID, TestTerminatedState.PARTIALLY_COMPLETE);
+          handleTestState(testID);
           break;
 
-        case GOAL_REACHABLE_REGRESSION_PREDICTION_ACCEPTABLE:
-        case GOAL_REACHABLE_NO_REGRESSION_EXPERIMENTS_EXECUTED:
-          testModelDAO.setTestState(testID, TERMINATED);
-          testModelDAO.setTestTerminatedState(testID, GOAL_REACHED);
+        case WAITING:
+        case START:
```
VincenzoFerme: How do you ensure that the execution of the task related to the `START` state does not interfere with the operation you have here?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
+          handleTestState(testID);
           break;
 
-        case GOAL_REACHABLE_REGRESSION_PREDICTION_NOT_ACCEPTABLE:
-          testModelDAO.setTestRunningState(testID, REMOVE_NON_REACHABLE_EXPERIMENTS);
+        case RUNNING:
+
+          TestRunningState runningState = testModelDAO.getTestRunningState(testID);
+
+          // set test to terminated
```
VincenzoFerme: First you need to handle the stop of executing tasks, then you can set the state to `TERMINATED`.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
+          testModelDAO.setTestState(testID, BenchFlowTestState.TERMINATED);
+          testModelDAO.setTestTerminatedState(testID, TestTerminatedState.PARTIALLY_COMPLETE);
+
+          // First we wait for any currently running tasks to finish since they finish quickly.
```
VincenzoFerme: How do you know they finish quickly? Think about computation of models...they can last some time.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
+          // Since we already set the test state to TERMINATED the task will not continue
```
VincenzoFerme: What do you do here? You should make it possible to interrupt the tasks, for example with a `cancel()` that gracefully stop the executing task (can also wait for its conclusion when it makes sense).

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
+              runningTaskFuture.get();
+            } catch (InterruptedException | ExecutionException e) {
+              // nothing to do
+            }
+          }
+
+          if (runningState == HANDLE_EXPERIMENT_RESULT
+              || runningState == DETERMINE_EXECUTE_VALIDATION_SET) {
+
+            // if an experiment is running we cancel it on the experiment-manager
```
VincenzoFerme: Why you need to schedule a task for this? I think that requesting the Experiment Manager to abort a given experiment is sufficient. If you use the task to reuse the code that is anyway used within a normal lifecycle (e.g., because both the test manager and the user can decide when to abort a task), then I would not schedule the task in the executor service, but simply execute it. Or tell me the specific reason why you use the executor service.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -221,281 +245,115 @@ private synchronized void handleTestRunningState(String testID) {
...
 
-      future.get();
-      testModelDAO.setTestRunningState(testID, nextState);
-      handleTestState(testID);
+      return benchFlowTestState == TERMINATED;
 
-    } catch (InterruptedException | ExecutionException e) {
-      // TODO - handle  properly
-      e.printStackTrace();
     } catch (BenchFlowTestIDDoesNotExistException e) {
```
VincenzoFerme, line 597: It is fine, but I would anyway keep the log message

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
```diff
@@ -0,0 +1,370 @@
...
+import java.util.concurrent.Future;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Jesper Findahl (jesper.findahl@gmail.com) created on 2017-07-02
+ */
+public class RunningStatesHandler {
+
+  /**
```
VincenzoFerme: If you gracefully stop the current scheduled and running tasks for the given `testID` that is going to be terminated, then this is not and should not be needed.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
```diff
@@ -0,0 +1,370 @@
...
+  public void determineExplorationStrategy(String testID) {
+
+    logger.info("determineExplorationStrategy with testID: " + testID);
+
+    DetermineExplorationStrategyTask task = new DetermineExplorationStrategyTask(testID);
+
+    // TODO - should go into a stateless queue (so that we can recover)
+    Future<?> future = taskExecutorService.submit(task);
+
+    // replace with new task
```
VincenzoFerme: Why replace? At any given point in time there should only be a single task running for a given `testID`, and the must be ensured.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
```diff
@@ -0,0 +1,370 @@
...
+
+    AddStoredKnowledgeTask task = new AddStoredKnowledgeTask(testID);
+
+    // TODO - should go into a stateless queue (so that we can recover)
+    Future<?> future = taskExecutorService.submit(task);
+
+    // replace with new task
+    testTasks.put(testID, future);
+
+    try {
```
VincenzoFerme, line 97: This can be handled by `waitForRunningTaskToComplete` accepting a callback function to be executed that can contain the following code.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
```diff
@@ -0,0 +1,370 @@
...
+
+      // wait for task to complete
+      future.get();
```
VincenzoFerme: If you make task cancellable,  then probably here you are going to get an [exception](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CancellationException.html) and you follow the exceptional flow, so that you do not need to check all the time if something is terminated. When you apply the previous comment, then the handling can be delegated to a method.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
```diff
@@ -0,0 +1,370 @@
...
+
+    DetermineExecuteExperimentsTask task = new DetermineExecuteExperimentsTask(testID);
+
+    // TODO - should go into a stateless queue (so that we can recover)
+    Future<?> future = taskExecutorService.submit(task);
+
+    // replace with new task
+    testTasks.put(testID, future);
+
+    // we don't wait for the task to complete since the experiment-manager
```
VincenzoFerme: Probably at this stage, the waiting for this state should be exactly to wait for the wanted data from the experiment manager, that should be the details for each of the trials that actually gets executed up to the point in which the experiment manager declares the experiment as completed in any of the possible states.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
```diff
@@ -0,0 +1,370 @@
...
+
+    // TODO - should go into a stateless queue (so that we can recover)
+    Future<Boolean> future = taskExecutorService.submit(task);
+
+    // replace with new task
+    testTasks.put(testID, future);
+
+    try {
+
+      // TODO - update: set next state as validate termination criteria
```
VincenzoFerme, line 197: You set it to `REMOVE_NON_REACHABLE_EXPERIMENTS`

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/tasks/abort/AbortRunningTestTask.java
```diff
@@ -0,0 +1,46 @@
...
+  public void run() {
+
+    logger.info("running: " + testID);
+
+    try {
+
+      String experimentID = testModelDAO.getRunningExperiment(testID);
+
+      if (experimentID != null) {
+        experimentManagerService.abortBenchFlowExperiment(experimentID);
```
VincenzoFerme, line 38: You might want to get and `ACK` about the execution of the operation, because otherwise it is really complex to control the behaviour of the distributed system.

benchflow-test-manager/application/src/test/java/cloud/benchflow/testmanager/scheduler/TestTaskSchedulerIT.java
```diff
@@ -267,6 +270,60 @@ public void runLoadTest() throws Exception {
 
   }
 
+  @Test
+  public void runBenchFlowTestTimeoutTest() throws Exception {
```
VincenzoFerme: You need also some tests where you submit multiple tests, and check the results of some of them, so that you can experience what happens with concurrency

VincenzoFerme (changes requested):

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/BenchFlowTestManagerApplication.java
jeppe-style: Why? I don't see any need to put this in the configuration

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: The running tasks cannot be cancelled - the only way is to wait for them to terminate. Therefore I need to check if the test has been terminated in the meantime. It is also what we discussed.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: I need to update the maxRunning time since that is read when the test goes back to `RUNNING`. In the `WAITING` state that is the only place I know how much time has elapsed since the `TimeoutTask` was scheduled.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: I don't understand the question fully

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: this happens in the `RUNNING` state meaning that is part of the execution.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: It actually happens also happens in the function in the next line. Although this wasn't changed in this PR.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: Not sure which case to handle. It doesn't matter if it already has been completed.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: I'm not sure I understand fully. It is only added in the `handleTestSate` method when the test is in state `READY`. This method is `synchronized` as well as the method here where it is removed.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: I think this just adds complexity when reading the code.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: we don't have any other state in the test life cycle.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: Each public method in this class is `synchronized`

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: I think this is what we afterwards discussed during the demo and is captured in this issue https://github.com/benchflow/benchflow/issues/499

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: I think this is what we afterwards discussed during the demo and is captured in this issue https://github.com/benchflow/benchflow/issues/499

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: It is not straightforward to cancel tasks because it means that in the task you always have to check if the task has been cancelled or not. If you think this change is necessary I think it is better to open an issue about it.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: It is because we don't want to keep the lock on the `TestTaskScheduler` longer than necessary.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: Which log message? And where to store it?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
jeppe-style: It is not straightforward to cancel tasks because it means that in the task you always have to check if the task has been cancelled or not. If you think this change is necessary I think it is better to open an issue about it.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
jeppe-style: This is just replacing in the data structure that keeps track of the currently running task for a given test. It is done throughout the code.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
jeppe-style: I don't see the benefit.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
jeppe-style: see previous comment about cancellable tasks.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
jeppe-style: this is what happens - the EM informs when the experiment has completed

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
jeppe-style: Not sure what the comment is about. This is a TODO

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/tasks/abort/AbortRunningTestTask.java
jeppe-style: I think this is covered in issue https://github.com/benchflow/benchflow/issues/499

benchflow-test-manager/application/src/test/java/cloud/benchflow/testmanager/scheduler/TestTaskSchedulerIT.java
jeppe-style: added that to the issue https://github.com/benchflow/benchflow/issues/499

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/BenchFlowTestManagerApplication.java
VincenzoFerme: Postponed in https://github.com/benchflow/benchflow/issues/517

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Fixed in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Fixed in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Fixed in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Fixed in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Postponed to https://github.com/benchflow/benchflow/issues/518

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Worked on this in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Worked on this in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: Worked on this in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
VincenzoFerme: I meant a log message in the `catch`, other than the ` e.printStackTrace();`, but it is fine like it is.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
VincenzoFerme: Worked on this in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0. It is fine to have.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
VincenzoFerme: Worked on this in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0. It is fine to have.

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/running/RunningStatesHandler.java
VincenzoFerme: Worked on this in https://github.com/benchflow/benchflow/pull/474#commits-pushed-08d26a0.

PR #480, opened 2017-07-06
Fix Bug with Creating Zip Files during Tests on Wercker Cloud

VincenzoFerme:
See title.

Fixes build error in https://github.com/benchflow/benchflow/pull/468: https://app.wercker.com/BenchFlow/benchflow/runs/build-benchflow-test-manager/5956d3513c2e15000193e386

PR #468, opened 2017-06-30
Adds support for single experiment test (load)

jeppe-style:
as per title

fixes https://github.com/benchflow/benchflow/issues/402

depends on PR https://github.com/benchflow/benchflow/pull/466

PR #481, opened 2017-07-06
Revert "Fix Bug with Creating Zip Files during Tests on Wercker Cloud"

VincenzoFerme:
Reverts benchflow/benchflow#480


VincenzoFerme:
I close this because it is not actually needed, since we can cleanup the history of https://github.com/benchflow/benchflow/pull/468

PR #479, opened 2017-07-05
Adds abort experiment API

jeppe-style:
as per title

See for review also https://github.com/benchflow/benchflow/issues/332#issuecomment-317889493

depends on PR https://github.com/benchflow/benchflow/pull/473

PR #484, opened 2017-07-06
Adds missing dependency in faban-client

jeppe-style:
as per title

depends on PR https://github.com/benchflow/benchflow/pull/479

### DIFF
--- a/benchflow-experiment-manager/application/pom.xml
+++ b/benchflow-experiment-manager/application/pom.xml
@@ -51,6 +51,12 @@
             <artifactId>httpmime</artifactId>
             <version>4.5.1</version>
         </dependency>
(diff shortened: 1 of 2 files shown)


VincenzoFerme:
@jeppe-style This fix should be done in the `faban-client`. Test with the most recent version you can find in https://github.com/benchflow/benchflow/pull/491, and if it works ping me so that we close this pull request and delete the branch.

VincenzoFerme (changes requested):
See https://github.com/benchflow/benchflow/pull/484#issuecomment-314215934

VincenzoFerme:
Proper fixed in https://github.com/benchflow/benchflow/pull/479/commits/0e0dd911465c29e4ae1037b3dceb10ef08a286d8

Issue #59, opened 2017-03-18
Update the DevOps Lifecycle


VincenzoFerme:
TODO

- it seems we can use only [Codacy](https://support.codacy.com/hc/en-us/articles/207994735) for all the projects. Test and document.

VincenzoFerme:
@igormoreno started to add a code checker for scala in https://github.com/benchflow/benchflow/pull/67. What we (@igormoreno) still need to do is: 

- improve the documentation to distinguish local lifecycle (that is described here currently: https://github.com/VincenzoFerme/benchflow-temporary-documentation/blob/master/development/contributing-to-benchflow.md) and CI lifecycle. The idea is that the developer should be aware of what he can do to check the code locally before committing it, and the checks should replicate exactly what he/she is going to get in CI. This is not related to a specific language or tool, but it is the general lifecycle that can be defined using a language specific build system (e.g., Maven for Java and Scala), and then @vincenzoferme will abstract it to an uniform interface for all the languages we have in benchflow.
- Use the issues you can find in codacy, in the Scala/Java projects, to fine tune the scalastyle configuration. In https://www.codacy.com/app/BenchFlow/benchflow/files?bid=4283320 you have a search bar to filter the issues and find the ones you are interested in.
- Description of the following commits have to go in the doc, in a subsection dedicated to Java/Scala projects for which the build system we use is Maven: https://github.com/benchflow/benchflow/pull/67/commits/aa4cfd013e9b06398553b01f89618cfaea0c9e70, https://github.com/benchflow/benchflow/pull/67/commits/04b7df158a7989f4ae35dbf04a9e85d77b51a65a
- work on https://github.com/benchflow/benchflow/pull/67#issuecomment-288514651
- document https://github.com/benchflow/benchflow/pull/111#issuecomment-290009775, and https://github.com/benchflow/benchflow/pull/111/commits/80d2ad9068fe42271466377e04135bab7940007c
- discuss why rebases when merging pull requests, and in general, are not the best thing to do
- decide about what to do according to what @jeppe-style is reporting in https://github.com/benchflow/benchflow/pull/67#issuecomment-288849885
- make sure the codacy settings added in the following pull request are actually working and applied: https://github.com/benchflow/benchflow/pull/237

VincenzoFerme:
[...]

PR #515, opened 2017-08-02
Fixes PMD Failures Failing the Builds

VincenzoFerme:
As per the title


VincenzoFerme:
I added the codify failure to the ignore list issue: #59

It is actually better to avoid import with `*`, but currently this is an overkill, because anyway the compiler takes care of optimising the code and importing what is needed by the code, and anyway the developer checks the code and the business logic, not what is imported.

PR #532, opened 2017-08-19
Adds terminating running state in test manager

### LINKED ISSUE #499: Test Life Cycle TERMINATING State
Add new state `TERMINATING` where we wait when a test is about to be terminated but we still need to get an update from the EM about the currently running experiment.

Before entering the `TERMINATING` state:
-  an `abort` of the currently running experiment (if any) is sent to the EM

Once in the state `TERMINATING`:
- waits for EM to notify that any running experiment has reached the state `TERMINATED`
- once notified the test is set as `TERMINATED`: `PARTIALLY COMPLETE` and handled as regular `TERMINATED` test

see updated life cycle diagram: https://github.com/VincenzoFerme/benchflow-temporary-documentation/blob/master/roadmap/Jesper/architecture/diagrams/svg/Test%20Life%20Cycle__BenchFlow%20Test%20Life%20Cycle_0.svg

jeppe-style:
as per title

fixes https://github.com/benchflow/benchflow/issues/499


VincenzoFerme:
@jeppe-style 

- you need to test that you go into the `TERMINATING` state. I do not see it. I need to go over a more careful review of this tomorrow.

- in #499 you write 

>waits for EM to notify that last experiment has reached state TERMINATED

How do you make sure that you wait for the **last** experiment?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -330,10 +332,11 @@ private void handleRunningTestState(String testID) throws BenchFlowTestIDDoesNot
         exit = true;
       }
       // Exit while waiting for the Experiment Manager to notify about the scheduled
-      // experiment to be executed.
+      // experiment to be executed or aborted (terminating state)
       else if (prevTestRunningState != null && testRunningState != null
```
VincenzoFerme: Are you sure this works? The point is that I check for the sequence of state in the current code, is you change taking into account this? Add some parenthesis as well.

jeppe-style:
- calling the `TERMINATING` state is basically happening [here](https://github.com/benchflow/benchflow/pull/532/files#diff-011215c1be2fbece1ba5058f98b5577cR437) because when a test is [terminated](https://github.com/benchflow/benchflow/pull/532/files#diff-4ee5285603cbdefec7bf569d0b6b6fb9R509) in the running state the next state is `TERMINATING`
- **last** experiment should maybe be reformulated to any **running** experiment has reached the state `TERMINATED`

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -480,14 +495,19 @@ public void terminateTest(String testID) {
...
               abortRunningTestTaskFuture.get();
+
             } catch (InterruptedException | ExecutionException e) {
               // nothing to do
               e.printStackTrace();
             }
 
           }
 
+          // set to TERMINATING and wait for experiment to terminate
```
VincenzoFerme: Probably this should happen before calling the `AbortRunningTask` because otherwise you could get the result from the `experiment-manager` before the switch to the `TERMINATING` state.

VincenzoFerme (changes requested):

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: good point

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: maybe we call about this?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -189,7 +190,8 @@ public synchronized void handleRunningTest(String testID) {
         }
         // Exit while waiting for the Experiment Manager to notify about the scheduled
         // experiment to be executed. This exits before the execution of HANDLE_EXPERIMENT_RESULT
-        else if (testRunningState != null && testRunningState.equals(HANDLE_EXPERIMENT_RESULT)) {
+        else if (testRunningState != null
+            && testRunningState.equals(TestRunningState.HANDLE_EXPERIMENT_RESULT)) {
```
jeppe-style: I think also here we need to check for `|| testRunningState == TestRunningState.TERMINATING`

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
jeppe-style: remove also `!= null` checks

jeppe-style: and also add a check that `RunninStatesHandler.terminating(...)` was called, for instance here https://github.com/benchflow/benchflow/blob/a19d162c400ca964b60158f6144b6baa901e2003/benchflow-test-manager/application/src/test/java/cloud/benchflow/testmanager/scheduler/TestTaskSchedulerIT.java#L378

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -480,14 +490,19 @@ public void terminateTest(String testID) {
...
+
             } catch (InterruptedException | ExecutionException e) {
               // nothing to do
               e.printStackTrace();
             }
 
           }
 
+          // run the TERMINATING state to terminate
+          handleRunningTest(testID);
```
jeppe-style: this should only be called in case we are not expecting a call from the EM

VincenzoFerme: And possibly from another iteration of the outer `while` loop, and not directly here

jeppe-style:
we should also check that `handleTerminatedState` is called when a test is terminated

VincenzoFerme:
@jeppe-style ready of in-progress?

benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -181,17 +191,19 @@ public synchronized void handleRunningTest(String testID) {
...
+        if (prevTestState == testState && (testState == WAITING || testState == TERMINATED)) {
           exit = true;
         }
         // Exit while waiting for the Experiment Manager to notify about the scheduled
         // experiment to be executed. This exits before the execution of HANDLE_EXPERIMENT_RESULT
-        else if (testRunningState != null
-            && testRunningState.equals(TestRunningState.HANDLE_EXPERIMENT_RESULT)) {
+        else if (testRunningState == TestRunningState.HANDLE_EXPERIMENT_RESULT
+            || testRunningState == TestRunningState.TERMINATING
+                && (prevTestRunningState == TestRunningState.HANDLE_EXPERIMENT_RESULT
```
VincenzoFerme: I do not get why you need to check this

jeppe-style: It is because in the case of `HANDLE_EXPERIMENT_RESULT` we should not execute the `TERMINATING` handler but wait for the trial to finish on Faban

VincenzoFerme: We should talk about this, it is faster

jeppe-style: add a comment

VincenzoFerme (approved):

PR #534, opened 2017-08-20
Improves experiment abort in Experiment Manager

### LINKED ISSUE #500: Improve aborting of experiment
When an experiment is running on Faban the Faban Manager polls for the result, however, with the current implementation it keeps the lock on the task scheduler and thus the experiment cannot be aborted. 

Solution
- add an API resource PUT `/result` on `/v1/users/{username}/tests/{testName}/{testNumber}/experiments/{experimentNumber}/trials/{trialNumber}` that takes the Faban result object as a parameter
  - when a request is received the `TaskScheduler` is passed the data (similar as in the TM when receiving an experiment result)
- Change the Faban Manager to call the API so that it easily can be changed into a separate service
- make tests to ensure that an experiment can be aborted during execution (waiting for Faban to complete execution of current trial)

### LINKED ISSUE #535: Change DropWizard Configuration Ports to 8080 and 8081
https://github.com/benchflow/benchflow/blob/ce7e83c7982dab048d96f08100a95310b71a875d/benchflow-experiment-manager/configuration.yml#L7-L14

jeppe-style:
as per title

- also changes FabanManager interaction to use a resource for submitting trial result

fixes https://github.com/benchflow/benchflow/issues/500
fixes https://github.com/benchflow/benchflow/issues/535

### DIFF
--- a/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
+++ b/benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
@@ -1,6 +1,7 @@
 package cloud.benchflow.experimentmanager.scheduler;
 
 import static cloud.benchflow.experimentmanager.models.BenchFlowExperimentModel.BenchFlowExperimentState.READY;
+import static cloud.benchflow.experimentmanager.models.BenchFlowExperimentModel.BenchFlowExperimentState.RUNNING;
 import static cloud.benchflow.experimentmanager.models.BenchFlowExperimentModel.BenchFlowExperimentState.TERMINATED;
 
 import cloud.benchflow.experimentmanager.BenchFlowExperimentManagerApplication;
@@ -96,30 +97,40 @@ public class ExperimentTaskScheduler {
    */
   public synchronized void handleStartingExperiment(String experimentID) {
 
-    boolean exit = false;
     BenchFlowExperimentState experimentState;
+    BenchFlowExperimentState prevExperimentState;
 
     try {
 
       experimentState = experimentModelDAO.getExperimentState(experimentID);
+      logger.info("handleStartingExperiment: experimentState == " + experimentState);
 
-      // Stop when we reach a final starting state
-      while (!exit) {
+      while (true) {
 
-        BenchFlowExperimentState prevTestState = experimentState;
+        prevExperimentState = experimentState;
 
         try {
-          experimentState = handleExperimentState(experimentID);
+          handleExperimentState(experimentID);
         } catch (BenchFlowExperimentIDDoesNotExistException e) {
           e.printStackTrace();
           break;
         }
 
-        // Exit as soon as ready or terminated is executed
+        experimentState = experimentModelDAO.getExperimentState(experimentID);
+        logger.info("handleStartingExperiment: (while) experimentState == " + experimentState);
+        logger.info(
+            "handleStartingExperiment: (while) prevExperimentState == " + prevExperimentState);
+
+        // Exit as soon as ready or terminated has been executed
         // An experiment can go to the TERMINATED state from START in case of errors
-        if (prevTestState == experimentState
-            && (experimentState == READY || experimentState == TERMINATED)) {
-          exit = true;
+        if (prevExperimentState == READY || prevExperimentState == TERMINATED) {
+          break;
+        }
+
+        // if the dispatcher changes the state before we exit this loop we also exit
+        // TODO - add test ensuring we always exit in this case
+        if (experimentState == RUNNING) {
+          break;
         }
 
       }
@@ -136,7 +147,6 @@ public class ExperimentTaskScheduler {
    */
   public synchronized void handleRunningExperiment(String experimentID) {
 
-    boolean exit = false;
     BenchFlowExperimentState experimentState;
     BenchFlowExperimentState prevExperimentState;
     RunningState runningState;
@@ -146,42 +156,37 @@ public class ExperimentTaskScheduler {
 
       experimentState = experimentModelDAO.getExperimentState(experimentID);
       runningState = experimentModelDAO.getRunningState(experimentID);
+      logger.info("handleRunningExperiment: experimentState == " + experimentState);
+      logger.info("handleRunningExperiment: runningState == " + runningState);
 
-      // Stop when we reach a final state
-      while (!exit) {
+      while (true) {
 
         prevExperimentState = experimentState;
         prevRunningState = runningState;
 
         try {
-          experimentState = handleExperimentState(experimentID);
+          handleExperimentState(experimentID);
         } catch (BenchFlowExperimentIDDoesNotExistException e) {
           e.printStackTrace();
           break;
         }
 
+        experimentState = experimentModelDAO.getExperimentState(experimentID);
         runningState = experimentModelDAO.getRunningState(experimentID);
-        logger.info("handleRunningExperiment: runningState == " + runningState);
-        logger.info("handleRunningExperiment: prevRunningState == " + prevRunningState);
+        logger.info("handleRunningExperiment: (while) experimentState == " + experimentState);
+        logger.info("handleRunningExperiment: (while) runningState == " + runningState);
+        logger.info("handleRunningExperiment: (while) prevRunningState == " + prevRunningState);
 
-        // Exit as soon as final state is executed
+        // Exit as soon as the final state has been executed
         if (prevExperimentState == TERMINATED) {
-          exit = true;
+          break;
         }
 
-        /*
-        Exit while waiting for the Faban Manager to notify about the result of the scheduled
-        trial. This exits before the execution of HANDLE_EXPERIMENT_RESULT
-        
-        NOTE for TERMINATING: Check depends if the system is waiting for input from the FabanManager
-        or not. So in the inner loop for running experiment we exit, but in the outer loop (HERE) we
-        check if we expect a result from the FabanManager or not. If we do not expect it we run the
-        loop again so that the TERMINATED state handling gets executed.
-        */
-        else if (runningState == RunningState.HANDLE_TRIAL_RESULT
-            || runningState == RunningState.TERMINATING
-                && prevRunningState == RunningState.HANDLE_TRIAL_RESULT) {
-          exit = true;
+
+        // Exit while waiting for the Faban Manager to notify about the result of the scheduled
+        // trial. This exits before the execution of HANDLE_EXPERIMENT_RESULT
+        if (runningState == RunningState.HANDLE_TRIAL_RESULT) {
+          break;
         }
 
       }
@@ -192,12 +197,12 @@ public class ExperimentTaskScheduler {
 
   }
 
-  private BenchFlowExperimentState handleExperimentState(String experimentID)
+  private void handleExperimentState(String experimentID)
       throws BenchFlowExperimentIDDoesNotExistException {
 
     BenchFlowExperimentState experimentState = experimentModelDAO.getExperimentState(experimentID);
 
-    logger.info("handleExperimentSate: " + experimentID + " state: " + experimentState.name());
+    logger.info("handleExperimentState: " + experimentID + " state: " + experimentState.name());
 
     switch (experimentState) {
       case START:
@@ -210,7 +215,7 @@ public class ExperimentTaskScheduler {
         break;
 
       case RUNNING:
-        handleRunningExperimentState(experimentID);
+        handleRunningState(experimentID);
         break;
 
       case TERMINATED:
@@ -222,7 +227,6 @@ public class ExperimentTaskScheduler {
         break;
     }
 
-    return experimentModelDAO.getExperimentState(experimentID);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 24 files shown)


benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -155,10 +158,19 @@ public synchronized void handleRunningExperiment(String experimentID) {
...
 
+        runningState = experimentModelDAO.getRunningState(experimentID);
+        logger.info("handleRunningExperiment: prevExperimentState == " + prevExperimentState);
+        logger.info("handleRunningExperiment: runningState == " + runningState);
+
         // Exit as soon as final state is executed
-        if (prevTestState != null && experimentState != null
-            && prevTestState.name().equals(experimentState.name())
-            && experimentState == TERMINATED) {
+        if (prevExperimentState == TERMINATED) {
```
VincenzoFerme: You can use `experimentState` because `TERMINATED` is a leaf state. You do it in https://github.com/benchflow/benchflow/pull/534/files#diff-e320626f292c5fd5e3bf5ea062ab6c90R285

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -265,8 +278,18 @@ private void handleRunningExperimentState(String experimentID)
...
+
       // Exit as soon as final state is executed
-      if (experimentState != null && experimentState == TERMINATED) {
+      if (experimentState == TERMINATED) {
+        exit = true;
+      }
+
+      // Exit while waiting for the Faban Manager to notify about the result of the scheduled
+      // trial. This exits before the execution of HANDLE_EXPERIMENT_RESULT
+      else if (runningState == RunningState.HANDLE_TRIAL_RESULT
```
VincenzoFerme: Why the code is different compared to https://github.com/benchflow/benchflow/pull/534/files#diff-e320626f292c5fd5e3bf5ea062ab6c90R171?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -349,15 +381,33 @@ public void abortExperiment(String experimentID) {
...
+          if (runningState != RunningState.TERMINATING) {
+
+            experimentModelDAO.setRunningState(experimentID, RunningState.TERMINATING);
+
+            // cancel the current running task, but let it complete before
+            cancelTask(experimentID);
+
+            // IF not waiting for Faban Manager we go into TERMINATING
+            if (runningState != RunningState.HANDLE_TRIAL_RESULT) {
+              handleRunningExperiment(experimentID);
```
VincenzoFerme: Why do you need this? And why can't this be handled by the normal lifecycle?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/faban/FabanManagerService.java
```diff
@@ -200,54 +202,75 @@ public RunId submitTrialToFaban(String experimentID, String trialID,
...
+        TrialResource trialResource = BenchFlowExperimentManagerApplication.getTrialResource();
+
+        TrialIDElements trialIDElements = new TrialIDElements(trialID);
+
+        trialResource.setFabanResult(trialIDElements.getUsername(), trialIDElements.getTestName(),
+            trialIDElements.getTestNumber(), trialIDElements.getExperimentNumber(),
+            trialIDElements.getTrialNumber(), fabanStatusRequest);
+      }
+
+    }).start();
```
VincenzoFerme, line 279: How do you ping back the experiment manager?

VincenzoFerme (changes requested):

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/faban/FabanManagerService.java
jeppe-style: it is there on line 266... basically I just call the resource directly without using a client

VincenzoFerme: Oh I see, I did not know that would work. And then where in the code the `while` loop handling the waiting will take back to go to the next state?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: yes right - even simpler

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: It depends if the system is waiting for input from the `FabanManager` or not. So in the inner loop for running experiment we exit, but in the outer loop we check if we expect a result from the `FabanManager` or not. If we do not expect it we run the loop again so that the `TERMINATED` state handling gets executed.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: We need to do this because we need to run the `TERMINATING` handler in case it is not triggered by the `FabanManger` sending the result of the trial.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/faban/FabanManagerService.java
jeppe-style: This code is executed as part of the `DETERMINE_EXECUTE_TRIALS`, and since it is in a thread it will be asynchronous, which will make the state of the experiment change to `HANDLE_TRIAL_RESULT` and the while loop will finish. Once the FabanManager has a result it will then invoke the Scheduler again and then the life cycle continues.

VincenzoFerme:
@jeppe-style it fails the build

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
VincenzoFerme: Ok, then update the comment with the description you provided here.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/faban/FabanManagerService.java
VincenzoFerme: Ok, then update the comment with the description you provided here.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: this doesn't work because we need to handle the `TERMINATED` state before exiting

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -96,30 +97,40 @@ protected void setRunningStatesHandler(RunningStatesHandler runningStatesHandler
...
-        if (prevTestState == experimentState
-            && (experimentState == READY || experimentState == TERMINATED)) {
-          exit = true;
+        if (prevExperimentState == READY || prevExperimentState == TERMINATED) {
+          break;
+        }
+
+        // if the dispatcher changes the state before we exit this loop we also exit
+        // TODO - add test ensuring we always exit in this case
+        if (experimentState == RUNNING) {
```
VincenzoFerme, line 132: This should not be possible

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -378,63 +332,72 @@ public void abortExperiment(String experimentID) {
...
+      cancelTask(experimentID);
 
-      switch (experimentState) {
+      /*
+      When we terminate an experiment it is important that we first cancel any running task and
+      then change the state to TERMINATED otherwise the state might be overwritten.
+      */
 
-        case START:
+      // lock to ensure execution is atomic
```
VincenzoFerme: > Only one thread can execute inside a Java code block synchronized on the same monitor object.

So it is not atomic, but accessed only by one Thread at a time. Other Threads acting on other part of the code can execute.

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
```diff
@@ -146,42 +156,37 @@ public synchronized void handleRunningExperiment(String experimentID) {
...
-        loop again so that the TERMINATED state handling gets executed.
-        */
-        else if (runningState == RunningState.HANDLE_TRIAL_RESULT
-            || runningState == RunningState.TERMINATING
-                && prevRunningState == RunningState.HANDLE_TRIAL_RESULT) {
-          exit = true;
+
+        // Exit while waiting for the Faban Manager to notify about the result of the scheduled
+        // trial. This exits before the execution of HANDLE_EXPERIMENT_RESULT
+        if (runningState == RunningState.HANDLE_TRIAL_RESULT) {
```
VincenzoFerme: What about the `TERMINATING` state now?

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: it was :-) that is why sometimes the trial went to the FAILURE because it was executing running in this loop and when there was no saved trial status it automatically said it was a failure.
If you look in the dispatcher you see that it just checks that there is something in the ready queue and if the running queue is empty it sets the state to running, but doesn't wait for this while loop to exit.

VincenzoFerme: Ok, then there is an issue in synchronisation between the `Dispatcher` and the `Scheduler`

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: it is not necessary anymore because we do the check when aborting https://github.com/benchflow/benchflow/pull/534/files/e07325058d8fb69b69981f8bf3470b6eb9f1c989..4723add973de6e9386fa66562e5649a44b3a640e#diff-e320626f292c5fd5e3bf5ea062ab6c90R383. 
The TERMINATING -> TERMINATED is then the same in all cases and it terminates once the state TERMINATED has been executed (line 181 above)

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: yes - it is practically atomic since the `handleStartingExperiment` and `handleRunningExperiment` are the only two other entry points and these methods are `synchronized`

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: yes I agree - however, in practice this works

jeppe-style: @jeppe-style open an issue with solution where dispatcher gets the lock of the scheduler object to ensure it executes after the while loop has terminated

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
VincenzoFerme: Could you please update the comment?

jeppe-style: @jeppe-style update comment accordingly

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/services/external/faban/FabanManagerService.java
jeppe-style: it was done in https://github.com/benchflow/benchflow/pull/534/commits/e07325058d8fb69b69981f8bf3470b6eb9f1c989

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: here is the issue https://github.com/benchflow/benchflow/issues/539

benchflow-experiment-manager/application/src/main/java/cloud/benchflow/experimentmanager/scheduler/ExperimentTaskScheduler.java
jeppe-style: comment added

VincenzoFerme (approved):

PR #538, opened 2017-08-22
Refactors and Simplifies Test Life Cycle

### LINKED ISSUE #533: All calls from a resource to task scheduler must be in a thread
Check all resources and ensure that all calls to the task scheduler are done in a thread, otherwise it will take too long before we can return to the caller.

### LINKED ISSUE #536: Add Minio Configuration Settings in the Test Manager as Well
Apply the following changes also in the Test Manager: https://github.com/benchflow/benchflow/pull/534/files/5fdad9e7d95dbeb5475eab603030f8d63caf0b78..4d4bc9156cf878a4df5fd53ab0954d6b70f556a7

jeppe-style:
as per title

also (in separate commits):
- fixes https://github.com/benchflow/benchflow/issues/533
- fixes https://github.com/benchflow/benchflow/issues/536

Note: It seems there is an issue that when a future is cancelled it is not recognized when checking so the life cycle continues. Therefore an additional check is made here https://github.com/benchflow/benchflow/pull/538/files#diff-4ee5285603cbdefec7bf569d0b6b6fb9R217

### DIFF
--- a/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
+++ b/benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
@@ -424,6 +424,9 @@ public class TestTaskScheduler {
       // cancel the current running task, but let it complete before
       cancelTask(testID);
 
+      // Lock to ensure execution is executed in an atomic fashion.
+      // It is practically atomic since the handleStartingExperiment and handleRunningExperiment
+      // are the only two other entry points and these methods are synchronized.
       synchronized (this) {
 
         BenchFlowTestState testState = testModelDAO.getTestState(testID);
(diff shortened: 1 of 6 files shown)


benchflow-test-manager/application/src/main/java/cloud/benchflow/testmanager/scheduler/TestTaskScheduler.java
```diff
@@ -174,48 +178,45 @@ public synchronized void handleRunningTest(String testID) {
...
+        // experiment to have terminated
+        if (testRunningState == TestRunningState.HANDLE_EXPERIMENT_RESULT
+            || testRunningState == TestRunningState.DETERMINE_EXECUTE_VALIDATION_SET) {
+          break;
+        }
+
+        // Exit if we are in a loop
+        // TODO - investigate if this can be removed and instead only use abort task
+        if (prevTestRunningState == testRunningState && testState != TERMINATED) {
+          break;
```
jeppe-style, line 219: @jeppe-style open an issue why this is here (e.g. explain also which test fails (testTaskSchedulerIT)) and link the code using shift

jeppe-style: here is the issue https://github.com/benchflow/benchflow/issues/540

VincenzoFerme (approved):

Issue #400, opened 2017-06-10
Initial semantic check DSL


jeppe-style:
**Configuration**
- [x] check that if goal should involve an `exploration` then `configuration.goal.explorationspace` is defined
- [x] check that if `configuration.users` is not specified then the goal is `exploration` and number of users is specified in the `configuration.goal.explorationspace` 
- [x] check that if goal is `load` then number of users is specified

**Workload**
- [x] check that at least one workload is defined
- [ ] check that operations and fixedSequenceMix names are the same

VincenzoFerme:
We also need to include checks on the infrastructure availability for the submitted specification. We need to decide if these are actually semantic checks or not.

PR #551, opened 2017-08-29
Adds initial semantic validation.

jeppe-style:
Asserts ifExplorationThenUsers, ifLoadGoalThenUsers, ifExplorationThenExplorationSpace and workloadIsDefined.

Fixes most issues in https://github.com/benchflow/benchflow/issues/400

### DIFF
--- a/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/BenchFlowTestAPI.scala
+++ b/benchflow-dsl/src/main/scala/cloud/benchflow/dsl/BenchFlowTestAPI.scala
@@ -3,7 +3,7 @@ package cloud.benchflow.dsl
 import cloud.benchflow.dsl.definition.BenchFlowTest
 import cloud.benchflow.dsl.definition.BenchFlowTestYamlProtocol._
 import cloud.benchflow.dsl.definition.errorhandling.{ BenchFlowDeserializationException, BenchFlowDeserializationExceptionMessage }
-import cloud.benchflow.dsl.definition.configuration.goal.goaltype.GoalType.LOAD
+import cloud.benchflow.dsl.definition.validation.SemanticValidation
 import net.jcazevedo.moultingyaml._
 
 import scala.util.{ Failure, Success, Try }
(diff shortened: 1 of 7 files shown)


VincenzoFerme:
@jeppe-style add also the missing check in #400 so we can close it

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/BenchFlowTestAPI.scala
```diff
@@ -50,11 +59,86 @@ object BenchFlowTestAPI {
 
   }
 
-  def validateTest(benchFlowTest: BenchFlowTest): Boolean = {
+  private val userMissing = "Missing user specification. "
```
VincenzoFerme: Move semantic validation to a different file, in a package where it is clear the responsibility of the code

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/BenchFlowTestAPI.scala
```diff
@@ -50,11 +59,86 @@ object BenchFlowTestAPI {
...
+
+  def validateTest(benchFlowTest: BenchFlowTest, validationErrorMessages: StringBuilder): Boolean = {
+
+    validationErrorMessages.append("Semantic validation failed. ")
+    var isValid = true
 
     // TODO - validate semantic in separate function on the object
 
-    false
+    // check that number of users is defined if exploration goal
```
VincenzoFerme: Is there a way to enable and execute just the checks that are needed, given the result of a previous check?

VincenzoFerme (changes requested):

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/BenchFlowTestAPI.scala
jeppe-style: updated in https://github.com/benchflow/benchflow/pull/551/commits/f666d1d36505922fcafa68ca138953785da65a9c

benchflow-dsl/src/main/scala/cloud/benchflow/dsl/BenchFlowTestAPI.scala
jeppe-style: You would then need to keep track of the errors and then execute just those checks. Why would you want to do that? Currently it is the user's responsibility to ensure errors are fixed.

PR #552, opened 2017-08-29
Updates DSL for TM and EM

jeppe-style:
as per title


VincenzoFerme:
Update after merging https://github.com/benchflow/benchflow/pull/551 to devel, so you can refer to devel.

